PR #2, opened 2021-06-01
Feat add vc implementation

niermann999:
Add the implementation of the algebra interface using vc library. The implementation is a vertical vectorization based on the existing array plugin.


asalzburger:
This was marked still WIP, please have a look

niermann999:
This is implementation works the way it is, but needs some makeover I think before it makes sense to merge, because it has some serious performance issues still. Changeing a tiny detail in one function erases them, but results in an obviously wrong calculation.

niermann999:
Tentative results: It seems the performance issues are systemload/hardware dependent.

asalzburger (approved):

asalzburger:
Thanks, even if not yet conclusive, I think we should merge that in.

PR #12, opened 2021-08-16
feat: Include vecmem for memory management

niermann999:
Includes vecmem for the algebra-plugins in preparation for a possible vecmem based plugin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,16 @@ endif()
 
 option(ALGEBRA_PLUGIN_BUILD_VC "Download and build local Vc" Off)
 
+if(ALGEBRA_PLUGIN_USE_VECMEM)
+     find_package( vecmem QUIET )
+
+     if( NOT vecmem_FOUND )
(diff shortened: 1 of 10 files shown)


beomki-yeo:
Thanks for the work. This looks good.
Can you also look into `core/include/array/include/algebra/definitions/array.hpp` to replace `std::array` with `array_s`?

niermann999:
Yes, I will address that next. I am thinking, the static array might get its own file or is it absolutely guaranteed that it will retain the exact behaviour of ```std::array```? The types you mean are actually more container types to the algebra types right now (e.g. for alignment)

beomki-yeo:
As far as I know std::array and vecmem::static_array works samely at least for algebras included in this repository. I am not sure if you have look into [algebra](https://github.com/acts-project/traccc/blob/main/core/include/definitions/algebra.hpp)  of traccc, but it simply replaced `std::array` with `vecmem::static_array`

asalzburger (approved):
Great to see this coming in!

PR #16, opened 2021-09-03
add rotate function to eigen

beomki-yeo:
Rotate function of eigen plugin is missing so I added it 
The input type (`Eigen::Transform<scalar, 3, Eigen::Affine>`) of eigen rotate function follows the type of `_data`
as the input type (`matrix44`) of array rotate function does. So we still keep the same usage for those plugins

@niermann999 @asalzburger would you review quickly as it is tiny PR?

### DIFF
--- a/core/include/eigen/include/algebra/definitions/eigen.hpp
+++ b/core/include/eigen/include/algebra/definitions/eigen.hpp
@@ -199,6 +199,18 @@ namespace algebra
                 return (_data.isApprox(rhs._data));
             }
 
+            /** Rotate a vector into / from a frame 
+             * 
+             * @param m is the rotation matrix
+             * @param v is the vector to be rotated
+             */
+	    
+            static vector3 rotate(const Eigen::Transform<scalar, 3, Eigen::Affine> &m, const vector3 &v)
+            {
+
+                return m.matrix().block<3,3>(0,0)*v;
+            }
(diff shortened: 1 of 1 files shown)


core/include/eigen/include/algebra/definitions/eigen.hpp
```diff
@@ -199,6 +199,18 @@ namespace algebra
                 return (_data.isApprox(rhs._data));
             }
 
+            /** Rotate a vector into / from a frame 
+             * 
+             * @param m is the rotation matrix
+             * @param v is the vector to be rotated
+             */
+	    
+            static vector3 rotate(const Eigen::Transform<scalar, 3, Eigen::Affine> &m, const vector3 &v)
```
niermann999, line 208: Maybe we can try this with the MatrixBase type and Eigen::Block<Derived>()? Then it might work with expressions, too

beomki-yeo: You mean changing `Eigen::Transform` into the type of `matrix44` in the argument? If so, that will lead into different argument semantic between eigen and other plugins. The argument type for `m` input should be same with the type of `_data`

niermann999: No, if this works, its fine, thank you!

PR #17, opened 2021-09-07
add inline to functions

beomki-yeo:
This PR adds inline qualifier to functions, to avoid duplicate definition errors and device code definition (w/ __host__ __device__ qualifier)


beomki-yeo:
@niermann999 Can you take a look?

niermann999:
Sorry, was on vacation

PR #19, opened 2021-09-15
fix: set build unittest to off

niermann999:
As this is generally a submodule, the unittests should only be build, when specifically requested


beomki-yeo:
Looks good to me; I wish hit approval button but I dont have write access :p

asalzburger (approved):

PR #20, opened 2021-09-16
use std::array for array_s 

beomki-yeo:
Currently `array_s` is an alias of `vecmem::static_array` when vecmem algebra is turned on. Unfortunately this doesn't work with detray because it cannot be called with `constexpr`: see [here](https://github.com/acts-project/detray/blob/8391a722efb2bc0e387f91ada2c4fff0b1f56a53/core/include/masks/cylinder3.hpp#L62) for example.

So let's roll it back to `std::array`

### DIFF
--- a/core/common/include/algebra/common/types.hpp
+++ b/core/common/include/algebra/common/types.hpp
@@ -24,19 +24,16 @@
 
 namespace algebra
 {
+    template <typename value_type, unsigned int kDIM>
+    using array_s = std::array<value_type, kDIM>;
+    
     // General algebra container types to be used with plugin
     #ifdef ALGEBRA_PLUGIN_USE_VECMEM
-
-    template <typename value_type, unsigned int kDIM>
-    using array_s = vecmem::static_array<value_type, kDIM>;
 
     template <typename value_type>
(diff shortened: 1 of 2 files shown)


niermann999:
Hm, I don't see a particular reason why this has to be a vecmem array type. As long as it is not dynamically allocated and the masks are stored in vecmem containers we should be fine. I will have a look

niermann999:
Yes, you are right. Lets put it in the detray: plugin/algebra/.../array_definitions.hpp for now, because that is easier to update between the projects. Or are there issues in traccc as well?

beomki-yeo:
you mean `plugins/algebra/include/detray/array/include/plugins/algebra/array_definitions.hpp`? 

how can it be done exactly? - you mean the following?

    template <typename value_type, unsigned int kDIM>
    using darray = std::array<value_type, kDIM>;
    
    template <typename value_type>
    using dvector = vecmem::vector<value_type>;

niermann999:
Yes, exactly. Because then we don't have to update the algebra-plugins submodule everywhere, right?

beomki-yeo:
Right. I updated acts-project/detray#87 would you have a look?

PR #22, opened 2021-09-23
avoid duplicate population of vecmem

beomki-yeo:
Currently in traccc, the compilation fails due to double population of vecmem (one from algebra plugin, and one from direct population) Let's avoid that


beomki-yeo:
@niermann999 Sorry for bothering you again - can you look into that? I also need to make another PR to update algebra plugins...

niermann999:
No problem! Lookk good

niermann999 (approved):

PR #29, opened 2021-10-19
avoid vecmem duplicate population

beomki-yeo:
w/o this, vecmem population may conflict with traccc


beomki-yeo:
Hmm okay. I think I should have looked into the new functionalities in the last PR. Thanks for suggestion though

PR #31, opened 2021-10-22
Code Rewrite, main branch (2021.10.22.)

krasznaa:
With a bit of a delay, this is the (actually!!) scary PR that I mentioned on Wednesday @niermann999.

It's still a work in progress, and the commits will probably need to be squashed eventually, but I wanted to give everybody a look at how I would imagine this project to work in the future. Though I imagine that I'll need to give a presentation about this, before it could be merged in.

The idea here is to properly separate in the code how data would be stored in memory by the users, and how calculations on that data would be done. The other basic idea is that I want to get rid of practically all `#ifdef`-s from the code. What I'm setting up here is a code organisation in which the type of an "algebra class/function" would clearly state
  - what type of objects it operates on;
  - what type of code it uses "internally" to perform its calculations.

The new structure of the code is the following:
  - I still kept the `algebra::common` library, though by now only a single header (`algebra/scalar.hpp`) is used from it, with a single definition. Later on I will want to get rid of this as well, but for now I decided to keep the compile-time decision about the scalar type.
  - I introduced the `storage/` directory to the repository, which holds headers/libraries declaring how data should be held in memory. The current libraries are:
    * `algebra::array_storage`: Using `std::array`, always enabled.
    * `algebra::eigen_storage`: Using `Eigen::Matrix`, enabled by `ALGEBRA_PLUGIN_INCLUDE_EIGEN`.
    * `algebra::smatrix_storage`: Using `ROOT::Math::SVector`, enabled by `ALGEBRA_PLUGIN_INCLUDE_SMATRIX`.
    * `algebra::vc_storage`: Using `Vc::SimdArray`, enabled by `ALGEBRA_PLUGIN_INCLUDE_VC`.
    * `algebra::vecmem_sorage`: Using `vecmem::static_array`, enabled by `ALGEBRA_PLUGIN_INCLUDE_VECMEM`.
  - I introduced the `math/` directory, which would hold all the "mathematics" with which data could be processed. The current libraries are:
    * `algebra::cmath_math`: Using "basic C\+\+" / `<cmath>`, always enabled.
    * `algebra::eigen_math`: Using `Eigen::Matrix`, enabled by `ALGEBRA_PLUGIN_INCLUDE_EIGEN`.
    * `algebra::smatrix_math`: Using `ROOT::Math::SMatrix`, enabled by `ALGEBRA_PLUGIN_INCLUDE_SMATRIX`.
    * `algebra::vc_math`: Using Vc vectorised operations, enabled by `ALGEBRA_PLUGIN_INCLUDE_VC`.
  - On top of these components I introduced the `frontend/` directory, which would hold libraries creating storage model - mathematics combinations. I will not list all of them here, have a look yourself. :wink:
  - For the tests I still kept `test_plugin.inl` with minimal modifications, even though I have ideas for making that a whole lot smarted as well. :stuck_out_tongue:
    * The change was that it now tests the "matrices" used by the transformations internally, for every "mathematics type". Not just the ones for which `__plugin_without_matrix_element_accessor` would not be declared.

As you can see, I generalised the heck out of `algebra::cmath_math`, to make it work on top of every storage model. I also taught `algebra::vc_math` to work with all of `algebra::array_storage`, `algebra::vecmem_storage`, and of course `algebra::vc_storage`. I'm still pondering about making `algebra::eigen_math` and `algebra::smatrix_math` capable of working with something other than their "own storage type", but that may not be super useful in the end.

So... Have a look, and let me know what you think. :wink:

Pinging @beomki-yeo, @stephenswat, @paulgessinger, @asalzburger.


math/vc/include/algebra/math/impl/vc_getter.hpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/** This method retrieves the norm of a vector, no dimension restriction
+ *
+ * @param v the input vector
+ **/
+template <typename vector_type>
+ALGEBRA_HOST_DEVICE inline auto norm(const vector_type &v) {
+
+  return std::sqrt(v[0] * v[0] + v[1] * v[1] + v[2] * v[2]);
```
niermann999, line 55: This is vc math, right? Why not use the vc dot product here?

math/vc/include/algebra/math/impl/vc_transform3.hpp
```diff
@@ -0,0 +1,409 @@
...
+
+namespace internal {
+
+/// 4x4 matrix type used by @c algebra::vc::math::transform3
+template <typename scalar_t>
+struct matrix44 {
+
+  /// Equality operator between two matrices
+  bool operator==(const matrix44 &rhs) const {
+    return ((x[0] == rhs.x[0]) && (x[1] == rhs.x[1]) && (x[2] == rhs.x[2]) &&
```
niermann999, line 30: This is not using explicit vectorization, when it could actually use the vc ```operator==``` and then a mask check ```isFull()```

math/vc/include/algebra/math/impl/vc_getter.hpp
krasznaa: Good catch. This must've fallen victim to one of my "unification experiments".

math/vc/include/algebra/math/impl/vc_transform3.hpp
krasznaa: I believe I borrowed this implementations from one of the other `transform3` classes as well. Since for the `std::array` case this equality was not explicitly declared. And I wonder if the default implementation of `==` for `std::array` is what we actually want here. (I'm honestly not sure.)

I'm 99% sure that the compiler would quite well auto-vectorise this expression, but you're absolutely right. I'll change this as well.

niermann999: if I remember correctly, there was an implementation in ```simd_types.hpp``` somewhere

krasznaa: For the Vc matrix type there is.

https://github.com/acts-project/algebra-plugins/blob/main/common/include/algebra/common/simd/simd_types.hpp#L102-L117

I think the equality being evaluated correctly was only ever a concern of mine for `std::array`. As all the other storage types should come with equality operators that behave like we would expect. And `std::array<std::array<FOO, 4>, 4>` may also do this. I just didn't trust that it would... (And not resort to pointer comparison for the outer-array.)

niermann999: Yes, thats the one I meant. It uses the Vc mask types for comparisons. Sorry for hiding it there :).

Ah, OK. I must admit, I don't know what ```std::equal``` would do.

krasznaa: Done.

krasznaa:
Hopefully these were all the comments... :wink: As if nothing else happens, I plan to merge this in by the end of today...

niermann999 (approved):

PR #33, opened 2021-11-02
Change auto into explicit size_type

beomki-yeo:
This PR is to fix the cuda application compilation error of acts-project/detray#134.

PR #34, opened 2021-11-02
Remove "auto" from algebra::cmath_math, main branch (2021.11.02.)

krasznaa:
Removed the `auto` size type specifications from the `algebra::cmath_math` code. As it was discovered by @beomki-yeo in https://github.com/acts-project/detray/pull/134, CUDA does not understand the formalism that the code used so far. :frowning:

All of the "free-standing" functions were updated to receive yet another template parameter, with the "size type". For this to work I had to make all of those functions make explicit use of the size type parameter. For instance turning:

```c++
template <template <typename, auto> class array_t, typename scalar_t>
ALGEBRA_HOST_DEVICE inline scalar_t dot(const array_t<scalar_t, 2> &a,
                                        const array_t<scalar_t, 2> &b) {
  return a[0] * b[0] + a[1] * b[1];
}
```

into:

```c++
template <typename size_type, template <typename, size_type> class array_t,
          typename scalar_t, size_type N, std::enable_if_t<N == 2, bool> = true>
ALGEBRA_HOST_DEVICE inline scalar_t dot(const array_t<scalar_t, N> &a,
                                        const array_t<scalar_t, N> &b) {
  return a[0] * b[0] + a[1] * b[1];
}
```

Otherwise the compiler would not be able to deduce the type of `size_type` automatically. Yeah, it's a bit weird, but it works. :stuck_out_tongue:

With these changes included I was able to build https://github.com/acts-project/detray/pull/134. While also keeping the SMatrix code functional. :wink: So if you're happy with it @beomki-yeo, this should replace #33...

### DIFF
--- a/frontend/array_cmath/include/algebra/array_cmath.hpp
+++ b/frontend/array_cmath/include/algebra/array_cmath.hpp
@@ -28,7 +28,7 @@ namespace array {
 /// @name cmath based transforms on @c algebra::array::storage_type
 /// @{
 
-using transform3 = cmath::transform3<array::storage_type, scalar>;
+using transform3 = cmath::transform3<std::size_t, array::storage_type, scalar>;
 using cartesian2 = cmath::cartesian2<transform3>;
 using polar2 = cmath::polar2<transform3>;
 using cylindrical2 = cmath::cylindrical2<transform3>;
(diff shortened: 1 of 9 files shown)


niermann999 (approved):

krasznaa:
Let's soldier on with this solution then. :smile:

PR #35, opened 2021-11-03
Test Rewrite, main branch (2021.11.03.)

krasznaa:
As I mentioned in #31, I had some ideas for re-writing the tests as well. :stuck_out_tongue:

In this PR I re-implemented the tests as GTest "typed tests".

https://github.com/google/googletest/blob/master/googletest/include/gtest/gtest-typed-test.h

Thereby removing the necessity of using "preprocessor tricks" for constructing the test code. Instead relying on GoogleTest's facilities to do the same. Which also allowed me to merge test sources into single executables where appropriate.

I don't actually claim that this formalism would be any simpler than what we had before. But it's certainly nicer C\+\+. Since the new `test_basics.hpp` header is an actual header. That can be included on its own as well. Kind of... But at least it does not depend on setting `__plugin` to work.


niermann999 (approved):
This will also be very useful in detray

PR #37, opened 2021-11-05
Externals and CMake Updates, main branch (2021.11.05.)

krasznaa:
These are largely some "mechanical" changes that I wanted to do for a while, and the update to [vecmem-0.6.0](https://github.com/acts-project/vecmem/releases/tag/v0.6.0) was a good excuse to make them.

First off, I replaced all `ALGEBRA_PLUGIN_<FOO>` cache variables with `ALGEBRA_PLUGINS_<FOO>` ones. It was triggering me from the start that the variables did not use the name of the project/repository. At the same time I also removed `ALGEBRA_PLUGIN_INCLUDE_ARRAY` from the README and the CI. As that variable doesn't exist since #31.

To make the setup of the externals more configurable, I introduced `ALGEBRA_PLUGINS_<FOO>_SOURCE` variables. These are marked as "advanced" variables, as they kind of are... :stuck_out_tongue: They can be used to tell the build where it should get the externals from. In case a client project wants to override some of these.

Note that Vc is still downloaded as a Git repository. Which is super annoying, as it takes ages to download even on a fast connection. But until a new numbered release is tagged with https://github.com/VcDevel/Vc/commit/d2b702f6500302ddeb21cdbef0b58a5551cff703 included in it, there's little else that we can do. :frowning:


niermann999 (approved):

niermann999:
On a different note, Iit would be great if we could update the detray README for a correct build config, too

PR #41, opened 2021-11-18
CI Update, main branch (2021.11.18.)

krasznaa:
As I mentioned in #39, we don't yet perform the CUDA tests in the CI.

I re-organised the CI configuration in the style that we use in [vecmem](https://github.com/acts-project/vecmem) these days.
  - I separately set up "native" and "containerized" tests;
  - The native tests turn on every feature that they can, and only test such a fully loaded configuration;
  - For the containerized tests, the "host tests" turn on features one-by-one, to test that those can function individually. And not just when everything else is also turned on.
  - The "cuda test", similar to the "native tests", only runs on one configurations, with everything possible turned on at once.

Note that I decided not to use every combination of build mode and scalar type in the tests individually. Since I intend to get rid of the `ALGEBRA_PLUGINS_CUSTOM_SCALARTYPE` setting completely soon, I didn't want to bloat the CI too much temporarily.

Also note, that I'm not actually super happy with how I set up the tests with the `ghcr.io/acts-project/ubuntu2004:v13` image. :frowning: I was hoping that I could make this type of formalism work:

```yaml
          - NAME: "HOST"
            CONTAINER: "ghcr.io/acts-project/ubuntu2004:v13"
            OPTIONS:
              - -DALGEBRA_PLUGINS_INCLUDE_EIGEN=TRUE -DALGEBRA_PLUGINS_SETUP_EIGEN3=TRUE -DALGEBRA_PLUGINS_USE_SYSTEM_EIGEN3=TRUE
              - -DALGEBRA_PLUGINS_INCLUDE_EIGEN=TRUE -DALGEBRA_PLUGINS_SETUP_EIGEN3=TRUE -DALGEBRA_PLUGINS_USE_SYSTEM_EIGEN3=FALSE
              ...
```

But GitHub did not want to play ball. :confused: So the configuration that I ended up with, does have some unnecessary-looking repetitions in it...

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -13,79 +13,113 @@ on:
       - main
       - 'release/**'
 
+# All the different build/test jobs.
 jobs:
-  ubuntu:
+
+  # Native build jobs.
+  native:
+
+    # The different build modes to test.
+    strategy:
+      matrix:
+        BUILD:
+          - TYPE: "Release"
+            SCALAR: "double"
+          - TYPE: "Debug"
+            SCALAR: "float"
+        PLATFORM:
+          - OS: "ubuntu-latest"
+            GENERATOR: "Unix Makefiles"
+          - OS: "macos-latest"
+            GENERATOR: "Xcode"
+
+    # The system to run on.
+    runs-on: ${{ matrix.PLATFORM.OS }}
+
+    # The build/test steps to execute.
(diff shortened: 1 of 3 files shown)


beomki-yeo:
Looks good to me. BTW what is your plan for `scalar_type` in the end?

beomki-yeo (approved):

PR #42, opened 2021-11-18
Windows Compatibility, main branch (2021.11.18.)

krasznaa:
These changes are set up on top of #41. So I'll keep it in a draft stage until that PR is sorted out.

I hope you all saw this coming. :stuck_out_tongue: I just really wanted to see if I could build this project on Windows as well. :smile: Turns out, we can. It only required a couple of changes. All of them in the test code, and not in the main headers / libraries of the project.

Note that unfortunately with CUDA enabled, I see a large number of warnings. Most of them coming from Eigen. But there are also some coming when `nvcc` is making use of [vecmem/memory/details/memory_resource_base.hpp](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/details/memory_resource_base.hpp). It seems that we'll need to find the equivalent of

```c++
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable : 4275)
#endif  // MSVC
```

for `nvcc`. :thinking: (Pinging @stephenswat and @paulgessinger.)

Still, the build succeeds, and all of the tests can run successfully on Windows. :smile:

One final thing that I didn't really understand is that the "recommended compile options" coming from Vc make the compilation (actually, the linking) fail. For non-Vc-using executables as well. :confused: (The linker fails to find standard library symbols for some reason with those flags enabled.) But I didn't want to start debugging that, so I just removed the global setup of the Vc flags with MSVC.

### DIFF
--- a/cmake/algebra-plugins-compiler-options.cmake
+++ b/cmake/algebra-plugins-compiler-options.cmake
@@ -4,6 +4,15 @@
 #
 # Mozilla Public License Version 2.0
 
+# Include the helper function(s).
+include( algebra-plugins-functions )
+
 # Set up the used C++ standard.
 set( CMAKE_CXX_STANDARD 17 CACHE STRING "The (host) C++ standard to use" )
 set( CMAKE_CUDA_STANDARD 17 CACHE STRING "The (CUDA) C++ standard to use" )
+
+# Turn on the correct setting for the __cplusplus macro with MSVC.
+if( "${CMAKE_CXX_COMPILER_ID}" MATCHES "MSVC" )
+   algebra_add_flag( CMAKE_CXX_FLAGS "/Zc:__cplusplus" )
+   algebra_add_flag( CMAKE_CUDA_FLAGS "-Xcompiler /Zc:__cplusplus" )
+endif()
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -32,6 +32,8 @@ jobs:
             GENERATOR: "Unix Makefiles"
           - OS: "macos-latest"
             GENERATOR: "Xcode"
+          - OS: "windows-latest"
+            GENERATOR: "Visual Studio 16 2019"
 
     # The system to run on.
     runs-on: ${{ matrix.PLATFORM.OS }}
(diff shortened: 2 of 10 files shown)


cmake/algebra-plugins-compiler-options.cmake
```diff
@@ -4,6 +4,15 @@
...
+
 # Set up the used C++ standard.
 set( CMAKE_CXX_STANDARD 17 CACHE STRING "The (host) C++ standard to use" )
 set( CMAKE_CUDA_STANDARD 17 CACHE STRING "The (CUDA) C++ standard to use" )
+
+# Turn on the correct setting for the __cplusplus macro with MSVC.
+if( "${CMAKE_CXX_COMPILER_ID}" MATCHES "MSVC" )
+   algebra_add_flag( CMAKE_CXX_FLAGS "/Zc:__cplusplus" )
+   algebra_add_flag( CMAKE_CUDA_FLAGS "-Xcompiler /Zc:__cplusplus" )
+endif()
```
krasznaa, line 18: A note to @stephenswat and @paulgessinger: I guess we should rather set up this flag (`/Zc:__cplusplus`) as a public compile flag on `vecmem::core`. :thinking: Since MSVC will always need to use this flag when looking at the headers from that library.

I'll play with this a bit, but something like

```cmake
target_compile_options( vecmem_core PUBLIC
   $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CXX>>, /Zc:__cplusplus>
   $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CUDA>>, -Xcompiler /Zc:__cplusplus> )
```

should work... :thinking:

krasznaa:
If nobody objects, I would just go ahead with this one. So that we can devote more time to #44. As that one I'll still need to test a lot more thoroughly.

niermann999:
Sorry, I have a slightly longer RTT currently. From my point of view go ahead, but I have to say I have no experience with windows whatsoever :)

niermann999 (approved):

krasznaa:
>  but I have to say I have no experience with windows whatsoever :)

Let's hope that it will not cause many failures in the CI going forward. As there was a bit of tension in the VecMem project when the CI started complaining about failures on Windows that @stephenswat and @paulgessinger couldn't easily test for themselves... :stuck_out_tongue:

PR #43, opened 2021-11-18
Remove CUDA vc_cmath Test, main branch (2021.11.18.)

krasznaa:
Removed the CUDA test for `algebra::vc_cmath`. That source file was added by mistake to the repository in #39. As I described in the PR, Vc does not work in device code. (The file was just a leftover from some tests in my source directory...)


krasznaa:
I'll just go ahead with this one...

PR #44, opened 2021-11-19
Remove algebra::scalar, main branch (2021.11.19.)

krasznaa:
Removed the `algebra::scalar` type definition. Instead making all of the code templated on the scalar type. Making the user choose the scalar type that they want to use, in "their code".

Updated the CUDA test code to produce more numerically sound results. But even with that, had to disable all fast math in order to achieve full agreement between the host and the device calculations at float precision.

I'm very open to suggestions on how this should be done exactly. As a first step [detray](https://github.com/acts-project/detray) will stay with its current setup. We will be able to pull in these definitions into appropriate namespaces for Detray, like:

```c++
namespace detray::algebra {
   using vector3 = algebra::array::vector3<float>;
}
```

Of course eventually I'll want to make the Detray templates also capable of receiving the scalar type in the final step of their construction.

While doing this, I renamed `algebra/common/algebra_qualifiers.hpp` into simply `algebra/qualifiers.hpp`. As we discussed this a long time ago with @beomki-yeo, now that all headers are under `algebra/`, the name of this file does not need to contain the name of the project. :wink:

Note that with this in place, the unit tests set up testing on top of `double` and `float` inside of the same executable. Finally making use of the infrastructure put in place in #35. :smile:

As a start this is still a draft, as I still have to update the CI for it. It also is not mergable without conflicts with #42. So we'll probably want to merge that one first, and then this one. (After having worked out the conflicts.)


niermann999:
In general I am fine with this :) I will fix the ```dscalar``` in the matplot++ soon, because that can only handle ```double``` anyway.

beomki-yeo (approved):

beomki-yeo:
Looks good to me!

krasznaa:
I believe it will be fine. However, I have not finished the migration of Detray to it completely in https://github.com/acts-project/detray/pull/167 yet.

Give me until tomorrow, and I'll spend some quality time on these two. Then we should be able to push both of the updates in...

krasznaa:
Ahh, never mind. I forgot about the details in the last 2 weeks... :stuck_out_tongue: Yes, this could go in as-is. https://github.com/acts-project/detray/pull/167 is ready to receive these changes as-is...

krasznaa:
Let me just make the executive decision...

PR #45, opened 2022-01-18
Improved SYCL Math Support, main branch (2022.01.18.)

krasznaa:
This -- in the end -- fairly substantial update was triggered by the errors seen in @konradkusiak97's developments in [traccc](https://github.com/acts-project/traccc). Which lead to the opening of the following ticket: https://github.com/intel/llvm/issues/5326

What we came across is that it would be impossible to use [std::atan2](https://en.cppreference.com/w/cpp/numeric/math/atan2) (and some other math functions) in SYCL device code on non-Intel backends. For the CUDA and HIP backends we must use `cl::sycl::atan2` instead.

For this I ended up introducing `algebra/math/common.hpp`, which would declare all of the "standard" math functions that we use in a couple of places in the code. This part was relatively simple.

But then I decided that we really also need to have explicit SYCL tests in place as well. Even if that test for now would only try to build the code for Intel backends. (For which the code was working as it was...) However once we introduce different SYCL backends for the [vecmem](https://github.com/acts-project/vecmem) CI, it should be relatively easy to introduce such tests here as well with this PR in place.

Pinging @beomki-yeo, @niermann999 and @HadrienG2. :wink:

### DIFF
--- a/.github/ci_setup.sh
+++ b/.github/ci_setup.sh
@@ -1,0 +1,22 @@
+# Algebra plugins library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+#
+# This script is meant to configure the build/runtime environment of the
+# Docker contaners that are used in the project's CI configuration.
+#
+# Usage: source .github/ci_setup.sh <platform name>
+#
+
+# The platform name.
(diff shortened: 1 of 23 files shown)


HadrienG2 (approved):
Seems reasonable to me.

niermann999 (approved):
Looks good

PR #47, opened 2022-02-04
Fix identity error in CUDA debug builds with Eigen

stephenswat:
This is a fix to a bug that @krasznaa alerted me of yesterday, where under some very strict set of circumstances, the Eigen tests would fail to run, giving an invalid memory access error. For this to happen, the code must be compiled in `Debug` build mode.

Anyway, the error results from the fact that the matrices in our Eigen `transform3` class are all defaulted to the `Identity()` transformation. While this should be fine, it apparently does not work well with CUDA. Thus, the solution is to remove these defaults and explicitly set the matrices to identity matrices using `setIdentity()`.

Without going deep into the Eigen3 code, I can only speculate _why_ this bug happens. One semi-likely explanation might be that the identity matrices are stored in the `.data` or `.bss` sections of the executable in order to allow direct copies, and that this memory is not available on the GPU. However, I am not sure whether this is the case.

### DIFF
--- a/math/eigen/include/algebra/math/impl/eigen_transform3.hpp
+++ b/math/eigen/include/algebra/math/impl/eigen_transform3.hpp
@@ -87,11 +87,9 @@ struct transform3 {
   /// @name Data objects
   /// @{
 
-  Eigen::Transform<scalar_type, 3, Eigen::Affine> _data =
-      Eigen::Transform<scalar_type, 3, Eigen::Affine>::Identity();
+  Eigen::Transform<scalar_type, 3, Eigen::Affine> _data;
 
-  Eigen::Transform<scalar_type, 3, Eigen::Affine> _data_inv =
-      Eigen::Transform<scalar_type, 3, Eigen::Affine>::Identity();
+  Eigen::Transform<scalar_type, 3, Eigen::Affine> _data_inv;
 
   /// @}
 
@@ -122,6 +120,8 @@ struct transform3 {
    **/
   ALGEBRA_HOST_DEVICE
   transform3(const vector3 &t) {
+    _data.setIdentity();
+
     auto &matrix = _data.matrix();
     matrix.template block<3, 1>(0, 3) = t;
 
@@ -153,8 +153,14 @@ struct transform3 {
     _data_inv = _data.inverse();
   }
 
+  /** Default constructor: set contents to identity matrices */
+  ALGEBRA_HOST_DEVICE
+  transform3() {
+    _data.setIdentity();
+    _data_inv.setIdentity();
+  }
+
   /** Default contructors */
-  transform3() = default;
   transform3(const transform3 &rhs) = default;
   ~transform3() = default;
 


math/eigen/include/algebra/math/impl/eigen_transform3.hpp
```diff
@@ -153,8 +153,14 @@ struct transform3 {
...
+  ALGEBRA_HOST_DEVICE
+  transform3() {
+    _data.setIdentity();
+    _data_inv.setIdentity();
+  };
+
   /** Default contructors */
-  transform3() = default;
   transform3(const transform3 &rhs) = default;
   ~transform3() = default;
```
krasznaa, line 165: Curiously enough it was these defaulted functions that caught my attention... After noticing them, I was convinced that they should be interfering with CUDA. But they don't.

Then I thought that aha, let's try adding `ALGEBRA_HOST_DEVICE` for the copy-constructor, add a line in one of the CUDA tests that copies such an object, and then we should get the same warning that Eigen treats us to on line:

https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/DenseStorage.h#L225

But that didn't happen either. :confused: So now I'm really not sure when CUDA likes and doesn't like functions to be declared `= default`. :confused:

In either case, the build also works with these defaulted functions not being declared at all. Since in https://github.com/acts-project/algebra-plugins/commit/62b6bb0ae58ece8d06ef12fa2958c99e9e3c38f2#diff-b0cb69d3ce20cba1d0bdd4c6bf0140e40433133e2c36bca83cc77444ab57603e I ended up doing just that, to get rid of a warning.

math/eigen/include/algebra/math/impl/eigen_transform3.hpp
```diff
@@ -153,8 +153,14 @@ struct transform3 {
     _data_inv = _data.inverse();
   }
 
+  /** Default constructor: set contents to identity matrices */
+  ALGEBRA_HOST_DEVICE
+  transform3() {
+    _data.setIdentity();
+    _data_inv.setIdentity();
+  };
```
krasznaa, line 161: Something is fishy with the warning configuration in this project. :confused: This semicolon really should've triggered one. :thinking: (I was removing such extra semicolons left and right from detray today.)

krasznaa (changes requested):
Apart from the one semicolon, this looks good. And I can confirm that it fixes the crash for me as well.

math/eigen/include/algebra/math/impl/eigen_transform3.hpp
krasznaa: Since I'm keen to make a new tag of the project for detray's benefit, I decided to be a bit aggressive, and removed the semicolon myself.

krasznaa (approved):
Let's go ahead with it as soon as the CI finishes.

math/eigen/include/algebra/math/impl/eigen_transform3.hpp
stephenswat: Sure, works for me. :smile:

PR #48, opened 2022-02-04
More MSVC Fixes, main branch (2022.02.04.)

krasznaa:
Fixed a warning coming from MSVC in `algebra::cmath::vector_getter`. It showed up while trying to build [detray](https://github.com/acts-project/detray) with MSVC. (Oh yeah, that is coming! MSVC complains about a lot of funky business in that project, for pretty good reasons...)

Added code for triggering the issue, so that this project would check that function explicitly as well. Even if the return value of that function is not checked. Because that return value is not independent of the plugin type being used...

Keeping it as a draft for now, as I'll only finalise it once detray builds successfully on Windows. :wink:


krasznaa:
I'll abandon trying to make detray work on Windows for now... It turns out that DFE would also need to be made compatible with MSVC first... 😦

So let's just take this one fix as-is.

PR #49, opened 2022-02-08
Define matrix type

beomki-yeo:
I am trying to implement the matrix which I defined in acts-project/detray#196

### DIFF
--- a/frontend/eigen_cmath/include/algebra/eigen_cmath.hpp
+++ b/frontend/eigen_cmath/include/algebra/eigen_cmath.hpp
@@ -74,6 +74,8 @@ template <typename T>
 using polar2 = cmath::polar2<transform3<T> >;
 template <typename T>
 using cylindrical2 = cmath::cylindrical2<transform3<T> >;
+template <typename T, std::size_t ROWS, std::size_t COLS>
+using matrix = math::matrix<matrix_type, T, ROWS, COLS>;
 
 /// @}
 
--- a/frontend/smatrix_cmath/include/algebra/smatrix_cmath.hpp
+++ b/frontend/smatrix_cmath/include/algebra/smatrix_cmath.hpp
@@ -70,6 +70,8 @@ template <typename T>
 using polar2 = cmath::polar2<transform3<T> >;
 template <typename T>
 using cylindrical2 = cmath::cylindrical2<transform3<T> >;
+template <typename T, std::size_t ROWS, std::size_t COLS>
+using matrix = math::matrix<matrix_type, T, ROWS, COLS>;
 
 /// @}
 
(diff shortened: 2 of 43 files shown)


frontend/eigen_cmath/include/algebra/eigen_cmath.hpp
```diff
@@ -74,6 +74,8 @@ template <typename T>
 using polar2 = cmath::polar2<transform3<T> >;
 template <typename T>
 using cylindrical2 = cmath::cylindrical2<transform3<T> >;
+template <typename T, std::size_t ROWS, std::size_t COLS>
+using matrix = math::matrix<matrix_type, T, ROWS, COLS>;
```
beomki-yeo, line 78: @krasznaa This is a bit broken since I used the same type from `eigen_eigen.hpp`..

frontend/smatrix_cmath/include/algebra/smatrix_cmath.hpp
```diff
@@ -70,6 +70,8 @@ template <typename T>
 using polar2 = cmath::polar2<transform3<T> >;
 template <typename T>
 using cylindrical2 = cmath::cylindrical2<transform3<T> >;
+template <typename T, std::size_t ROWS, std::size_t COLS>
+using matrix = math::matrix<matrix_type, T, ROWS, COLS>;
```
beomki-yeo, line 74: same here

frontend/vc_cmath/include/algebra/vc_cmath.hpp
```diff
@@ -88,6 +88,8 @@ template <typename T>
 using polar2 = cmath::polar2<transform3<T> >;
 template <typename T>
 using cylindrical2 = cmath::cylindrical2<transform3<T> >;
+template <typename T, std::size_t ROWS, std::size_t COLS>
+using matrix = math::matrix<matrix_type, T, ROWS, COLS>;
```
beomki-yeo, line 92: same here

krasznaa:
If I understand the situation correctly, you "just" need to be able to declare matrix objects in a generic way, while being able to read/write elements of the matrix with a uniform interface.

The code already does something like that, since `algebra::cmath::transform3` needed to be able to interact with different kinds of matrices.

https://github.com/acts-project/algebra-plugins/blob/main/math/cmath/include/algebra/math/impl/cmath_transform3.hpp#L21-L22

Unfortunately I can't promise that I would be able to do it today, but let me play with this a bit myself. (I should find time for this tomorrow afternoon.) I think it would be possible to re-purpose existing code for doing this. Otherwise I'll just make comments on your changes. (I'll have a few. :stuck_out_tongue:)

PR #50, opened 2022-02-11
Add Matrix Type, main branch (2022.02.11.)

krasznaa:
@beomki-yeo, this is my take on how we should introduce the `matrix_type` type into the repository.

First off, I introduced `algebra::array::matrix_type`, `algebra::eigen::matrix_type`, etc. in pretty much the same way as you did in #49.

But I didn't create a type called `matrix`. Instead, very much how we have `algebra::getter::vector(...)` since the beginning, I introduced a new set of `algebra::getter::element(...)` functions, which could be used to retrieve elements of a matrix.

And I didn't just implement those `algebra::getter::element(...)` functions from scratch. As I wrote in #49, we already have a set of `element_getter` functors in the code. Those are used by `algebra::cmath::transform3`, so that it could work on top of any matrix type. What I did now was to remove some duplications from the `element_getter` definitions, and to start using a single implementation of those functors both in `algebra::cmath::transform3`, and for the newly introduced `algebra::getter::element(...)` functions.

Then it came time to introduce tests for these new things. And that was the thing that took the most time on this. But I'm happy to report that the functions are now tested to work on the host, with CUDA and with SYCL as well.

I had to realise that I didn't test the build of the code with CUDA and/or SYCL on Windows in quite a while. The CUDA build can be made to work, although even that is super fragile. Since CUDA issues a separate list of warnings on Windows than it does on Linux. And I just can't find the right incantation for silencing more than one warning at a time from the CMake configuration. :frowning: But at the very least, if `Werror` is not used, the code does build with CUDA on Windows.

However SYCL is a completely different story. As it turns out, the very latest version of DPC\+\+ is unable to use `std::array` inside of kernels on Windows. :astonished: I will be opening a ticket to Intel about that next week.

So, have a look, I'm curious what you think!

### DIFF
--- a/frontend/array_cmath/include/algebra/array_cmath.hpp
+++ b/frontend/array_cmath/include/algebra/array_cmath.hpp
@@ -63,25 +63,12 @@ ALGEBRA_HOST_DEVICE inline array::storage_type<scalar_t, SIZE> vector(
                               SIZE>()(m, row, col);
 }
 
-/// Function extracting an element from a matrix (const)
-template <typename scalar_t, std::size_t ROWS, std::size_t COLS>
-ALGEBRA_HOST_DEVICE inline scalar_t element(
-    const array::matrix_type<scalar_t, ROWS, COLS>& m, std::size_t row,
-    std::size_t col) {
+/// @name Getter functions on @c algebra::array::matrix_type
+/// @{
 
-  return cmath::element_getter<std::size_t, array::storage_type, scalar_t>()(
-      m, row, col);
-}
+using cmath::element;
 
-/// Function extracting an element from a matrix (non-const)
-template <typename scalar_t, std::size_t ROWS, std::size_t COLS>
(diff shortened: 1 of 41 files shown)


krasznaa:
:thinking: The `algebra::getter::vector` functions I had to put into the "frontend libraries", as the matrix types used by the `transform3` classes actually differ between for instance `vc_cmath` and `vc_vc`.

But you're right, for the `algebra::getter::element` function the situation is a bit different. I have an idea how to set this up in a nice way, I'll make that change.

Note that the "matrix operations" may very well be put into a helper class/struct in a future PR. I just didn't think that the act of accessing an element of a matrix warranted a separate class. But for the operations that we need, we'll need to think a bit about their best implementation.

beomki-yeo:
Good. I think its good to go.

> But for the operations that we need, we'll need to think a bit about their best implementation.

Right maybe we can start with very simple operations (e.g. transpose).

PR #57, opened 2022-03-30
Add set block function in matrix

beomki-yeo:
This is possibly the last update for the moment ( I hope ... )

This PR adds `set_block` function to matrix backend.
Eigen backend doesn't need `set_block` as `block` supports lvalue.
Meanwhile, `block` operator of cmath and smtrix backends don't support lvalue so I needed to setup a new function

Yeah so it is an awful implementation and might degrade the performance. We need to think about how to make all backends support lvalue block operation. Or, at least find a better alternative

Following snippet is what I want to achieve in the end

```c++
matrix<4,4> m;
auto b22 = matrix_actor().template block<2,2>(m,1,1); // get 2x2 block matrix which can be used as lvalue
element_getter()(b22,0,0) = 1; //  m(1,1) is also changed
```

### DIFF
--- a/math/cmath/include/algebra/math/impl/cmath_matrix.hpp
+++ b/math/cmath/include/algebra/math/impl/cmath_matrix.hpp
@@ -64,6 +64,18 @@ struct actor {
     return block_getter().template operator()<ROWS, COLS>(m, row, col);
   }
 
+  /// Operator setting a block
+  template <size_type ROWS, size_type COLS, class input_matrix_type>
+  ALGEBRA_HOST_DEVICE void set_block(input_matrix_type &m,
+                                     const matrix_type<ROWS, COLS> &b, int row,
+                                     int col) {
+    for (size_type i = 0; i < ROWS; ++i) {
+      for (size_type j = 0; j < COLS; ++j) {
+        element_getter()(m, i + row, j + col) = element_getter()(b, i, j);
+      }
+    }
+  }
+
   // Create zero matrix
   template <size_type ROWS, size_type COLS>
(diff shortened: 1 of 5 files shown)


niermann999:
Please ignore this comment if I misunderstand your intention, but is the idea that ```m(1,1) is also changed``` not rather a side-effect in this circumstance? Maybe we should call it a block view to make it clear

beomki-yeo:
I am trying to make it same with Eigen 
Please look into this: [Eigen Block operation](https://eigen.tuxfamily.org/dox/group__TutorialBlockOperations.html)

I don't like calling it `block_view` to be honest

beomki-yeo:
And following is how the block is used in ACTS covariance transport

https://github.com/acts-project/acts/blob/8016339ffb9d1f10bc623f0d12df1a7279560c51/Core/include/Acts/Propagator/detail/GenericDefaultExtension.hpp#L138-L233

niermann999:
It doesn't bother me in particular, I just wanted to remark that I find it confusing :)  And it seems that also in ACTS the operation got commented on this

beomki-yeo:
Then do you prefer manually calling `set_block` instead of getting references?

beomki-yeo:
I didn't push this into the PR yet - cmath block can be lvalue by defining block as `array<array<scalar*,ROWS>,COLS>`
The real difficulty comes from SMatrix whose block doesn't support lvalue

beomki-yeo:
I hope so - yeah but let us get by with `set_block` for the moment. 
Implementing lvalue brings too many artifacts which I don't want to handle right now...

PR #60, opened 2022-04-01
Add table to README.md

paulgessinger:
Mirroring what we have in traccc.


beomki-yeo:
Nice. Actually the rest parts of README are also pretty outdated.

PR #68, opened 2022-06-10
Add a constructor for transform3

beomki-yeo:
This PR adds a constructor of `transform3` which takes `t`,`x`,`y`,`z` and `get_inverse` as inputs. The last argument is to decide whether to calculate inverse or not. (I have a use case for creating tf3 object without inverse)
The old constructor also utilizes the new constructor to avoid code repetition.

### DIFF
--- a/math/cmath/include/algebra/math/impl/cmath_transform3.hpp
+++ b/math/cmath/include/algebra/math/impl/cmath_transform3.hpp
@@ -60,8 +60,8 @@ struct transform3 {
   /// @name Data objects
   /// @{
 
-  matrix44 _data;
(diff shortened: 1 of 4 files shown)


niermann999 (approved):
The methods that use the inverse matrix will for some plugins produce random results when the inverse is not properly initialized, or am I wrong? I know it is not used currently, but uninitialized vectors in the vc plugin might be a bad idea in itself

beomki-yeo:
I just made `transform3` (except vc) have identity matrix as a default. I have no idea how to do it for `vc_transform3` - let's just leave it as it is for now

PR #69, opened 2022-06-18
Added Google Benchmark to the project.

wermos:
This PR adds Google Benchmark, as a first step to having some benchmarking on the math functions defined in this repository so that we can quantitatively measure the performance of the functions and classes defined in here.

I did my best to mimic the existing project structure, and took inspiration from [VecMem](https://github.com/acts-project/vecmem/) where appropriate.


wermos:
Requesting @niermann999 for a review

wermos:
Also, I was unsure of how we want the structure of the code inside the `benchmarks` directory to look, so I have left it blank for now. Any suggestions are welcome.

CMakeLists.txt
```diff
@@ -84,6 +86,20 @@ if( ALGEBRA_PLUGINS_SETUP_GOOGLETEST )
    endif()
 endif()
 
+# Set up Google Benchmark
+option( ALGEBRA_PLUGINS_SETUP_GOOGLEBENCHMARK
```
beomki-yeo, line 90: Let's call it `BENCHMARK` rather than `GOOGLEBENCHMARK` as we do for detray.

extern/googlebenchmark/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# Algebra plugins library, part of the ACTS project (R&D line)
+#
+# (c) 2021 CERN for the benefit of the ACTS project
```
beomki-yeo, line 3: 2022

CMakeLists.txt
```diff
@@ -84,6 +86,20 @@ if( ALGEBRA_PLUGINS_SETUP_GOOGLETEST )
...
+   "Set up the Google Benchmark target(s) explicitly" TRUE )
+option( ALGEBRA_PLUGINS_USE_SYSTEM_GOOGLEBENCHMARK
+   "Pick up an existing installation of Google Benchmark from the build environment"
+   FALSE )
+if( ALGEBRA_PLUGINS_SETUP_GOOGLEBENCHMARK )
+   if( ALGEBRA_PLUGINS_USE_SYSTEM_GOOGLEBENCHMARK )
+      find_package( benchmark REQUIRED )
+   else()
+      add_subdirectory( extern/googlebenchmark )
```
beomki-yeo, line 99: let's also name it `benchmark`

CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ option( ALGEBRA_PLUGINS_INCLUDE_VECMEM
    "Include VecMem types in Algebra Plugins" FALSE )
 option( ALGEBRA_PLUGINS_BUILD_TESTING "Build the unit tests of Algebra Plugins"
    TRUE )
+option( ALGEBRA_PLUGINS_BUILD_BENCHMARKS "Build the benchmark suite of Algebra Plugins"
```
beomki-yeo, line 49: I don't think we need this for now because `TRUE` option will make an error. It will be okay to add this option once you have something in `benchmarks` directory.

beomki-yeo (commented):
Looks good to me. The best way to implement benchmark would be reusing the existing infrastructure in `tests/common` directory.

extern/googlebenchmark/CMakeLists.txt
wermos: Ah, good catch.

CMakeLists.txt
wermos: I was actually copying VecMem here, where they called it [`GOOGLEBENCHMARK`](https://github.com/acts-project/vecmem/blob/df310ef85a4ec74234f43e18ccbf472c70ba9e8e/benchmarks/CMakeLists.txt#L9). Nevertheless, I have changed it to be more in line with the `algebra-plugins` way of doing things.

extern/googlebenchmark/CMakeLists.txt
wermos: Here, I actually copied the beginning portion (that is, the copyright stuff) from the [`extern/googletest`](https://github.com/acts-project/algebra-plugins/blob/d94ef468209a0c6eadb67130bf50ca37a19a40e7/extern/googletest/CMakeLists.txt#L3) CMakeLists.txt. It has 2021 there too.

Should I change it in this PR, or make a new PR for it? It looks to me like there are quite a few places which have the copyright line still on 2021.

CMakeLists.txt
wermos: I plan on adding some tests in the next PR itself. Plus, the default option (for now, at least), is `FALSE`. However, if you think it will cause problems (maybe in the CI or something), then I can remove it and add it into the next PR.

extern/googlebenchmark/CMakeLists.txt
niermann999: You can change it in this PR. In general, when you add a new file, the copyright should be set to the year it is deployed. If you modify old code in existing files, the copyright will be extended to the year it was modified e.g. ```2021-2022```.

niermann999 (approved):

CMakeLists.txt
beomki-yeo: You can leave it if you like

PR #70, opened 2022-07-08
Common Coordinate Transform

beomki-yeo:
This PR Integrates the coordinate transform (cylindrical2, polar2, cartesian2) for different backends.


krasznaa (changes requested):
I do not like this. :frowning: You introduce "header files" with this PR that can not be included directly. They can only be included through "pre-processor trickery". :frowning: (So they shouldn't even have the `.hpp` extension, as that's just bound to lead to confusion.)

Are you trying to cut down on the code duplication in the repository? I can get on board with that. But we should do it in a "proper C\+\+ way". :thinking: Though I'm not entirely sure what that should be...

PR #72, opened 2022-08-10
local to global position transformation in coordinate structs

beomki-yeo:
PR depends on #71 

This PR adds local to global transformations in coordinates, which follows ACTS conventions. I also added `line2` coordinate.


beomki-yeo:
@niermann999 I added `cartesian3` and `cylindrical3`

tests/common/test_host_basics.hpp
```diff
@@ -534,6 +534,227 @@ TYPED_TEST_P(test_host_basics, local_transformations) {
   ASSERT_NEAR(polfrom2[1], polfrom3[1], this->m_epsilon);
 }
 
+// This test cartesian2 coordinate
+TYPED_TEST_P(test_host_basics, cartesian2) {
+
+  // Preparation work
+  const typename TypeParam::vector3 z =
+      typename TypeParam::vector_actor().normalize(
+          typename TypeParam::vector3{0., 0., 1.});
```
niermann999, line 543: Aren't these already normalized?

tests/common/test_host_basics.hpp
```diff
@@ -534,6 +534,227 @@ TYPED_TEST_P(test_host_basics, local_transformations) {
...
+    typename TypeParam::scalar r = 0.;
+    typename TypeParam::scalar radius() const { return r; }
+  };
+
+  const typename TypeParam::scalar r = 2.;
+  const cylinder_mask mask{r};
+
+  // Global position on surface
+  const typename TypeParam::point3 global1 = {
+      typename TypeParam::scalar{3.4142136},
```
niermann999, line 662: I am not sure using ```scalar``` explicitly will gain much in this context, since it is still a ```scalar``` constructed from a ```double``` regardless, if I am not mistaken? Also, this is just as verbose as using ```static_cast```

math/common/include/algebra/coordinates/line2.hpp
```diff
@@ -0,0 +1,92 @@
...
+    const point3 t = trf.translation();
+
+    // Radial vector
+    const vector3 r = vector_actor().cross(z, d);
+
+    // Assign the sign depending on the position w.r.t line
+    // Right: -1
+    // Left: 1
+    const scalar_type sign =
+        vector_actor().dot(r, t - p) > 0. ? scalar_type{-1.} : scalar_type{1.};
```
niermann999, line 68: I would put the this as ```vector_actor().dot(r, t - p) > scalar_type{0} ? -1 : 1;``` to avoid a potential single vs double precision comparison, but please someone correct me if I am not making sense. Or use ```static_cast<scalar_type>(0.)``` maybe, if you want to write it with ```double``` literals?

tests/common/test_host_basics.hpp
```diff
@@ -534,6 +534,227 @@ TYPED_TEST_P(test_host_basics, local_transformations) {
...
+      typename TypeParam::scalar{4.4142136}, typename TypeParam::scalar{9.}};
+
+  // Global to local transformation
+  const typename TypeParam::point2 local = c2.global_to_local(trf, global1);
+
+  // Check if the local position is correct
+  ASSERT_NEAR(local[0], r * M_PI / 4., this->m_isclose);
```
niermann999, line 669: And then here you are using ```doubles``` all the way. I am not sure how much consistency we want to require for this. Is this actually important or am I mostly pedantic in the testing code?

niermann999 (changes requested):
Thanks a lot for adding the new types!

math/common/include/algebra/coordinates/line2.hpp
```diff
@@ -0,0 +1,92 @@
...
+  inline point2 operator()(const point3 &local3, scalar_type sign) const {
+
+    return {sign * vector_actor().perp(local3), local3[2]};
+  }
+
+  /** This method transform from a point from global cartesian 3D frame to a
+   * local 2D line point */
+  ALGEBRA_HOST_DEVICE
+  inline point2 global_to_local(const transform3_type &trf, const point3 &p,
+                                const vector3 &d) const {
```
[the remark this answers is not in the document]
beomki-yeo: `d` is the direction of the particle, and `z` of trf is the direction of the line.
We need the direction of the track to get the sign (left or right)

tests/common/test_host_basics.hpp
beomki-yeo: Right

math/common/include/algebra/coordinates/coordinate_base.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+// Project include(s).
+#include "algebra/qualifiers.hpp"
+
+namespace algebra {
+
+/** Coordinate base struct
+ */
+template <typename transform3_t>
+struct coordinate_base {
```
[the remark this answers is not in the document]
beomki-yeo: Yes. It will be extended in the future when I add jacobian transformation.

tests/common/test_host_basics.hpp
beomki-yeo: You are right. I will remove the scalar cast

math/common/include/algebra/coordinates/line2.hpp
beomki-yeo: Fixed it as suggested.

tests/common/test_host_basics.hpp
beomki-yeo: I think it is OK as I don't see any narrowing conversion warnings

niermann999: As far as I know, you will not receive many narrowing warnings on implicit conversions without using curly braces. In this case, however, I guess it will not give you a warning anyway. I don't think that comparisons will trigger a warning (or whatever google test is doing)?

beomki-yeo: OK. I put the scalar conversion to `M_PI_4` now. Please tell me explicitly if it needs to be fixed.

tests/common/test_host_basics.hpp
beomki-yeo: Looks like we need the conversion here. Otherwise, windows compiler fails which I don't know why

tests/common/test_host_basics.hpp
niermann999: I am not quite following, but if you want to be on the safe side, I guess you could static cast the ```double``` to a ```scalar_type```. That this still implicitly casts ```r``` to ```double``` doesn't matter so much in tests I would say
```
static_cast<scalar_type>(r * M_PI / 4.)
```

niermann999: The test should work fine anyway, so, I'll leave it up to you

niermann999 (approved):

PR #73, opened 2022-08-21
Roll back to v0.13.0 and Remove coordinates 

beomki-yeo:
As we discussed in the last meeting, I removed coordinates and I will add them into detray instead.

Following that, I rolled back all the changes that I made (#72 and #71) to enable common coordinate libraries.

I also added the enumeration of `track_indices` for bound and free track parameters for different plugins. This is because all plugins have different matrix index (cmath: std::size, eigen: int, smatrix: unsigned int). They will be used as template parameter in detray.

UPDATE on Aug 22th:
As we discussed today, I removed the plugin-specfic track indices.


niermann999:
Hm, can we maybe split the rollback and the track indices or is that not possible?

beomki-yeo:
track index was added first in this PR and removed in the later commit, so you don't have to check out the changes in track index which doesn't exist.
It would make sense to split the roll back and coordinate removal though. But it would be quite tough to split them at the moment...

PR #77, opened 2022-10-10
Added Fastor

wermos:
I've added Fastor as a math backend to the repository. This is step one to adding it as a full-fledged math backend, similar to Eigen and SMatrix.


niermann999:
If there are no objections, let's merge this

PR #80, opened 2022-11-20
Partial Pivot LU decomposition for matrix inversion

beomki-yeo:
This PR is motivated by the poor precision of the cofactor method of `cmath` plugin: The single precision result of the cofactor method is quite different from the double precision one.

Following values is the determinant of 6x6 matrix that appears during the kalman filtering

- eigen-eigen float: 0.00000000004306401069
- eigen-eigen double: 0.00000000004306355415

- **array-cmath float with cofactor: 0.00000000004327766964 (BAD)**
- array-cmath double with cofactor: 0.00000000004306355415

- array-cmath float with partial pivot LU decomposition: 0.00000000004306400722
- array-cmath double with partial pivot LU decomposition: 0.00000000004306355415 

It is obvious that LU decomposition gives more reliable result with single precision.

### DIFF
--- a/frontend/array_cmath/include/algebra/array_cmath.hpp
+++ b/frontend/array_cmath/include/algebra/array_cmath.hpp
@@ -99,11 +99,10 @@ template <typename scalar_t, class... As>
 using actor =
     cmath::matrix::determinant::actor<size_type, matrix_type, scalar_t, As...>;
 
-// determinant::cofactor
+// determinant::partial_pivot_lud
 template <typename scalar_t, size_type... Ds>
-using cofactor =
-    cmath::matrix::determinant::cofactor<size_type, matrix_type, scalar_t,
-                                         element_getter<scalar_t>, Ds...>;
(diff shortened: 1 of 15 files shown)


krasznaa (approved):
I have to admit the math of the `partial_pivot_lud` classes went way over my head. :frowning:

From a technical perspective the changes look good to me. But just to expose my ignorance: Clients will not be expected to use types like `algebra::determinant::partial_pivot_lud` directly, right? Because if they are allowed to do that, we will need to implement these types for the remaining frontends (like `eigen_eigen`) as well.

But all in all, I need to trust your judgement on this...

beomki-yeo:
>  Clients will not be expected to use types like algebra::determinant::partial_pivot_lud directly, right? 

Right. They should not be exposed to clients in general. 
In case users want to call the decomposed matrix thru `algebra::cmath::matrix::decomposition::lud`, it will be problematic because SMatrix doesn't seem to have LU decompose function.

But I don't think we have a specific use case for that.

PR #85, opened 2023-01-23
Eigen getter size type template parameter

beomki-yeo:
As we've discussed today...

### DIFF
--- a/math/eigen/include/algebra/math/impl/eigen_getter.hpp
+++ b/math/eigen/include/algebra/math/impl/eigen_getter.hpp
@@ -96,12 +96,14 @@ ALGEBRA_HOST_DEVICE inline auto eta(
 /// Functor used to access elements of Eigen matrices
 struct element_getter {
   /// Get non-const access to a matrix element
-  template <
-      typename derived_type, typename size_type_1, typename size_type_2,
-      std::enable_if_t<std::is_base_of<Eigen::DenseCoeffsBase<
-                                           derived_type, Eigen::WriteAccessors>,
-                                       Eigen::MatrixBase<derived_type> >::value,
-                       bool> = true>
+  template <typename derived_type, typename size_type_1, typename size_type_2,
+            std::enable_if_t<
+                std::is_base_of<
+                    Eigen::DenseCoeffsBase<derived_type, Eigen::WriteAccessors>,
+                    Eigen::MatrixBase<derived_type> >::value &&
+                    std::is_convertible<size_type_1, Eigen::Index>::value &&
+                    std::is_convertible<size_type_2, Eigen::Index>::value,
+                bool> = true>
   ALGEBRA_HOST_DEVICE inline auto &operator()(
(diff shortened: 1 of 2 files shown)


stephenswat:
I think having two separate type variables for the rows and columns is a bit excessive; maybe we can just use the same type for both?

math/eigen/include/algebra/math/impl/eigen_getter.hpp
```diff
@@ -97,53 +97,58 @@ ALGEBRA_HOST_DEVICE inline auto eta(
 struct element_getter {
   /// Get non-const access to a matrix element
   template <
-      typename derived_type,
+      typename derived_type, typename size_type_1, typename size_type_2,
       std::enable_if_t<std::is_base_of<Eigen::DenseCoeffsBase<
                                            derived_type, Eigen::WriteAccessors>,
                                        Eigen::MatrixBase<derived_type> >::value,
                        bool> = true>
```
krasznaa, line 104: I'm just throwing it out there, I'm not absolutely convinced myself that it's a good idea. But... should we add an explicit check that the received type is convertible to `int`? Like:

```c++
      std::enable_if_t<std::is_base_of<Eigen::DenseCoeffsBase<
                                           derived_type, Eigen::WriteAccessors>,
                                       Eigen::MatrixBase<derived_type> >::value &&
                       std::is_convertible<size_type_1, int>::value &&
                       std::is_convertible<size_type_2, int>::value,
                       bool> = true
```

Unfortunately, in case of giving bad arguments to the function, the compilation errors will be hard to decipher either way.  Adding these statements would be more for the human reader of the code I guess.

krasznaa (commented):
With all of this said... Isn't the access operator in question itself templated? :thinking:

https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/plugins/IndexedViewMethods.h

(The Eigen code is complicated enough that I'm a little confused myself though.)

As such, shouldn't it be able to handle these type differences gracefully? Can you point me at a concrete warning coming from the Eigen code with the latest macOS clang compiler?

math/eigen/include/algebra/math/impl/eigen_getter.hpp
beomki-yeo: OK

beomki-yeo:
@krasznaa Here is the example of errors from clang. It is not from macOS clang but it's the same error anyway..
```
/home/beomki/projects/detray/detray/tests/common/include/tests/common/coordinate_cartesian2.inl:98:45: error: implicit conversion loses integer precision: 'std::size_t' (aka 'unsigned long') to 'int' [-Werror,-Wshorten-64-to-32]
                EXPECT_NEAR(m.element(J, i, j), 1., isclose);
```

krasznaa (approved):
I like it. I was just about to say that the `static_cast`-s are not actually needed, but you removed those yourself as well. :wink:

PR #90, opened 2023-02-03
Eigen SYCL Tests, main branch (2023.02.03.)

krasznaa:
Introduced SYCL tests for the `algebra::eigen_cmath` and `algebra::eigen_eigen` frontends.

After a wild goose chase with [Eigen](https://eigen.tuxfamily.org) in
https://github.com/intel/llvm/issues/8065, I now added these tests with SYCL finally. So that if Eigen does start to cause problems, we could catch it.

Note that this test was left out "back in the day" because I didn't know yet that one needed to use `EIGEN_NO_CUDA`, `EIGEN_NO_HIP` and also `EIGEN_DONT_VECTORIZE` to prevent Eigen from doing stuff inside of SYCL kernels that it shouldn't. I learned all of this while working with the [traccc](https://github.com/acts-project/traccc) code...


tests/accelerator/sycl/CMakeLists.txt
```diff
@@ -19,11 +19,23 @@ target_link_libraries( algebra_tests_sycl_common
...
+if( ALGEBRA_PLUGINS_INCLUDE_EIGEN )
+   algebra_add_test( eigen_sycl "eigen_cmath.sycl" "eigen_eigen.sycl"
+      LINK_LIBRARIES algebra::eigen_cmath algebra::eigen_eigen
+                     algebra::tests_sycl_common GTest::gtest_main )
+   # Prevent Eigen from considering the SYCL device compilation as either:
+   #  - host compilation;
+   #  - CUDA compilation;
+   #  - HIP compilation.
+   target_compile_definitions( algebra_test_eigen_sycl
+      PRIVATE EIGEN_NO_CUDA EIGEN_NO_HIP EIGEN_DONT_VECTORIZE )
```
beomki-yeo, line 36: Does "NO CUDA compilation" imply that we are not supposed to use SYCL with cuda backend?

krasznaa: Quite the opposite. :smile:

When building this test with the Intel compiler's CUDA backend, the preprocessor macros in Eigen think that they are being processed by `nvcc`. And try to make use of symbols that LLVM doesn't know about.

With `EIGEN_NO_CUDA` we tell Eigen not to use CUDA specific formalisms, and just rely on standard C\+\+.

`EIGEN_DON_VECTORIZE` is very similar. If one compiles for the `spir64_x86_64` backend at the same time as for any of the GPU backends, Eigen tries to use vectorization intrinsics in the code. Which of course the device code compiler freaks out about.

So yeah, Eigen still has ways to go for convenient SYCL usage. :stuck_out_tongue: (Though the phase space with this is so huge that I wonder if it even makes sense to try to automate all of this behind the scenes...)

krasznaa: P.S. I am able to run these tests with the latest "Intel compiler nightly" from a single test binary on OpenCL, CUDA and HIP backends at the same time. :wink:

beomki-yeo (approved):

PR #95, opened 2023-03-13
feat: SoA layout for vector

niermann999:
This adds a new vector wrapper class that allows for an SoA layout, in particular, for the ```Vc::Vector``` type. Also adds benchmarks to compare to the ```std::array``` backend


niermann999:
This is still VERY WIP

niermann999:
Since this PR is already huge, the SoA layout for ```std::array``` will be done in a separate PR

.gitignore
```diff
@@ -30,3 +30,7 @@
 *.exe
 *.out
 *.app
+
+# Local folders
+/bin
+/build
```
krasznaa, line 36: Wait... why do we have such a monster of a `.gitignore` file to begin with? :confused: 

I can get on board with adding `build/`. While I don't do it, I know that strange people like @stephenswat (:stuck_out_tongue:) like to put that into the source directory. But what about `bin/`? And what about all those other silliness-es that were already in the file? :confused:

benchmarks/common/include/benchmark/common/benchmark_getter.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "benchmark_vector.hpp"
+
+namespace algebra::bench_op {
+
+// Functions to be benchmarked
+
+struct phi {
+  inline static const std::string name{"phi"};
+  template <typename vector_t>
+  decltype(auto) operator()(const vector_t &a) const {
```
krasznaa, line 20: I still have a deep distrust for `decltype(auto)`. We could just use `auto` here as well, no?

math/vc_soa/include/algebra/math/impl/vc_soa_getter.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+/// This method retrieves phi from a vector, vector base with rows >= 2
+///
+/// @tparam N dimension of the vector
+/// @tparam scalar_t scalar type
+/// @tparam array_t array type that holds the vector elements
+///
+/// @param v the input vector
+template <std::size_t N, typename scalar_t,
+          template <typename, std::size_t> class array_t,
```
krasznaa, line 37: In the past it hurt us a couple of times when we hardcoded the "size variable" of an array type to be `std::size_t`. (Since some libraries use `unsigned int` instead.) Is that not expected to become a problem here? :thinking:

krasznaa (commented):
There are **tons** that I don't understand in here... :frowning: Could you possibly post a more detailed explanation of the goals into the PR? SoA is of course good. I just don't see yet how this code here is really meant to work... :confused:

benchmarks/common/include/benchmark/common/benchmark_getter.hpp
niermann999: Sure

.gitignore
niermann999: Let me remove ```bin/```. I leave the rest for you to decide :)

niermann999:
OK, I added a bit more of an explanation :)

math/vc_soa/include/algebra/math/impl/vc_soa_getter.hpp
niermann999: And even with ```Vc::SimdArray``` it needs an alias type which seems to confuse the MSVC compiler maybe? I am open to suggestions here

math/cmath/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ algebra_add_library(algebra_cmath_math cmath_math
    "include/algebra/math/impl/cmath_operators.hpp"
    "include/algebra/math/impl/cmath_transform3.hpp"
    "include/algebra/math/impl/cmath_vector.hpp"
+   "include/algebra/math/impl/cmath_vector_soa.hpp"
```
krasznaa, line 16: I don't see this file in this PR. :confused: (:thinking: we may want to make it an error when we point at non-existent files with CMake...)

math/vc_soa/include/algebra/math/vc_soa.hpp
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+// clang-format off
+#include "algebra/math/impl/vc_soa_vector.hpp"
+#include "algebra/math/impl/vc_soa_getter.hpp"
```
krasznaa, line 13: It's not great that the two would have a fixed order for being included. :frowning: Why is that?

storage/common/include/algebra/storage/array_operators.hpp
```diff
@@ -0,0 +1,107 @@
...
+/// @{
+template <template <typename, std::size_t> class array_t, typename scalar_t,
+          std::size_t N,
+          std::enable_if_t<!std::is_same_v<array_t<scalar_t, N>, scalar_t>,
+                           bool> = true>
+ALGEBRA_HOST_DEVICE inline constexpr array_t<scalar_t, N> operator*(
+    const array_t<scalar_t, N> &a, scalar_t s) noexcept {
+
+  array_t<scalar_t, N> result;
+  for (std::size_t i{0u}; i < N; ++i) {
```
krasznaa, line 28: I don't think I've ever seen such a formalism in a for-loop. :confused: I would really prefer to stay with the classical `std::size_t i = 0u` formalism here...

storage/common/include/algebra/storage/vector.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+  /// Holds the data value for every vector element
+  array_t<value_t, N> m_data;
+
+ private:
+  /// Sets the trailing uninitialized values to zero.
+  template <std::size_t... Is>
+  constexpr void zero_fill(std::index_sequence<Is...>) noexcept {
+    if constexpr (sizeof...(Is) > 0) {
+      //((m_data[N - sizeof...(Is) + Is] = value_t(0)), ...);
```
krasznaa, line 130: ¿que?

krasznaa (commented):
Are the issues on macOS really insurmountable? I really don't like how that code is disabled right now. :frowning:

At the same time, let's discuss this face to face at some point. I'd love to understand better where this thing is meant to be heading.

storage/common/include/algebra/storage/vector.hpp
niermann999: This is meant to fill the rest of the vector with zeros in case too few arguments are provided (like ```vc_array4``` does). Problem being that Vc seems to interpret the comma operator as some sort of initialization overload on its ```Vc::Vector``` type and warns that this is deprecated, which becomes an error in the CI. Something fishy is going on here, but I only need it for the ```Vc::SimdArray``` based plugin later and I don't want to come up with the thing from scratch again, so I commented it for now. Open to suggestions here :)

math/vc_soa/include/algebra/math/vc_soa.hpp
niermann999: I need the dot product in the getter functionality, I just included the header now

benchmarks/common/include/benchmark/common/benchmark_getter.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+// Functions to be benchmarked
+
+struct phi {
+  inline static const std::string name{"phi"};
+  template <typename vector_t>
+  auto operator()(const vector_t &a) const {
+    return algebra::getter::phi(a);
+  }
+};
```
wermos, line 23: For these, maybe defining a macro and then instantiating a bunch of those macro instances would be a cleaner way of doing this?

storage/common/include/algebra/storage/vector.hpp
```diff
@@ -0,0 +1,243 @@
...
+  /// Default contructor sets all entries to zero.
+  constexpr vector() { zero_fill(std::make_index_sequence<N>{}); }
+
+  /// Construct from element values @param vals .
+  ///
+  /// In order to avoid uninitialized values, which deteriorate the performance
+  /// in explicitely vectorized code, the underlying data array is filled with
+  /// zeroes if too few arguments are given.
+  template <typename... Values>
+  constexpr vector(Values &&... vals) : m_data{vals...} {
```
krasznaa, line 41: @niermann999, I played with this a little. MSVC becomes okay with this line if you give up on the rvalue reference, and go with a simple lvalue one. I.e.

```suggestion
  constexpr vector(const Values &... vals) : m_data{vals...} {
```

I'm anyway not sure why rvalue formalism was chosen here in the first place. :confused:

krasznaa: :thinking: Or at least #97 starts working that way... I was doing my tests in that branch, not this one...

niermann999: Great, thanks a lot for having a look! It's a universal reference, so it can be an rvalue or an lvalue. I was hoping to circumvent copies when passing in larger objects, like e.g. `Vc::Vector` or `std::array`

krasznaa: Right... I vaguely remember having issues with universal references in MSVC in the past... 🤔 To be honest, the templating here is complicated enough that I can't say at this stage if MSVC is right to complain or not. 😦

I think, if copy and move constructors are important to have here, we would need to declare them separately. 😦

krasznaa:
P.S. I saw some "new" Eigen warnings from the CUDA build in my tests. Something to follow up on separately at one point. (Those may be only showing up on Windows, didn't test it rigorously.)

storage/common/include/algebra/storage/vector.hpp
stephenswat: Maybe `m_data{std::forward<Values>(vals)...}` will stop MSVC from complaining?

krasznaa: I was absolutely convinced that Stephen would be right. But alas, no... That's not the silver bullet after all. 😦

Having played with the code a bit more, the whole shebang gets triggered from line 32 in `benchmark_vector.hpp`

```c++
  std::vector<vector_t> a, b, results;
```

```
C:\Users\krasz\ATLAS\algebra\algebra-plugins\benchmarks\common\include\benchmark\common\benchmark_vector.hpp(32,1): message : see reference to class template instantiation 'std::vector<vector_t,std::allocator<vector_t>>' being compiled [C:\Users\krasz\ATLAS\algebra\build\benchmarks\algebra_benchmark_vc_soa_getter.vcxproj]
          with
          [
              vector_t=algebra::storage::vector<3,Vc_1::Vector<double,Vc_1::VectorAbi::Sse>,algebra::vc_soa::storage_type>
          ]
```

Clearly MSVC wants to use that very generic constructor as the copy/move constructor for the class. And fails at doing so.

Having experimented with a number of different SFINAE ideas on this, trying to only enable this constructor when the arguments are "just so", in the end I gave up. And made these changes instead:

```diff
diff --git a/math/vc_soa/include/algebra/math/impl/vc_soa_vector.hpp b/math/vc_soa/include/algebra/math/impl/vc_soa_vector.hpp
index b47a4a9..b640ab2 100644
--- a/math/vc_soa/include/algebra/math/impl/vc_soa_vector.hpp
+++ b/math/vc_soa/include/algebra/math/impl/vc_soa_vector.hpp
@@ -39,8 +39,8 @@ ALGEBRA_HOST_DEVICE inline storage::vector<N, Vc::Vector<scalar_t>, array_t>
 cross(const storage::vector<N, Vc::Vector<scalar_t>, array_t> &a,
       const storage::vector<N, Vc::Vector<scalar_t>, array_t> &b) {

-  return {Vc::fma(a[1], b[2], -b[1] * a[2]), Vc::fma(a[2], b[0], -b[2] * a[0]),
-          Vc::fma(a[0], b[1], -b[0] * a[1])};
+  return {{Vc::fma(a[1], b[2], -b[1] * a[2]), Vc::fma(a[2], b[0], -b[2] * a[0]),
+           Vc::fma(a[0], b[1], -b[0] * a[1])}};
 }

 /// Dot product between two input vectors
diff --git a/storage/common/include/algebra/storage/vector.hpp b/storage/common/include/algebra/storage/vector.hpp
index 4f0681a..784c3ff 100644
--- a/storage/common/include/algebra/storage/vector.hpp
+++ b/storage/common/include/algebra/storage/vector.hpp
@@ -32,21 +32,10 @@ class vector {
   /// Default contructor sets all entries to zero.
   constexpr vector() { zero_fill(std::make_index_sequence<N>{}); }

-  /// Construct from element values @param vals .
-  ///
-  /// In order to avoid uninitialized values, which deteriorate the performance
-  /// in explicitely vectorized code, the underlying data array is filled with
-  /// zeroes if too few arguments are given.
-  template <typename... Values>
-  constexpr vector(Values &&... vals) : m_data{vals...} {
-    if constexpr ((sizeof...(Values) < N) &&
-                  (!std::conjunction_v<std::is_same<array_type, Values>...>)) {
-      zero_fill(std::make_index_sequence<N - sizeof...(Values)>{});
-    }
-  }
-
   /// Construct from existing array storage @param vals .
   constexpr vector(const array_type &vals) : m_data{vals} {}
+  /// Construct from existing array storage @param vals .
+  constexpr vector(array_type &&vals) : m_data{vals} {}

   /// Assignment operator from wrapped data.
   ///
diff --git a/tests/vc_soa/vc_soa.cpp b/tests/vc_soa/vc_soa.cpp
index d81c421..fbf9030 100644
--- a/tests/vc_soa/vc_soa.cpp
+++ b/tests/vc_soa/vc_soa.cpp
@@ -27,8 +27,8 @@ TEST(test_vc_host, vc_soa_vector) {
   // Value type is Vc::Vector<float>
   using value_t = typename vector3_v::value_type;

-  vector3_v a{value_t::One(), 2.f * value_t::One(), 3.f * value_t::One()};
-  vector3_v b{4.f * value_t::One(), 5.f * value_t::One(), 6.f * value_t::One()};
+  vector3_v a{{value_t::One(), 2.f * value_t::One(), 3.f * value_t::One()}};
+  vector3_v b{{4.f * value_t::One(), 5.f * value_t::One(), 6.f * value_t::One()}};

   EXPECT_TRUE((a[0] == value_t(1.f)).isFull());
   EXPECT_TRUE((a[1] == value_t(2.f)).isFull());
@@ -111,7 +111,7 @@ TEST(test_vc_host, vc_soa_getter) {
   // Value type is Vc::Vector<float>
   using value_t = typename vector3_v::value_type;

-  vector3_v a{value_t::One(), 2.f * value_t::One(), 3.f * value_t::One()};
+  vector3_v a{{value_t::One(), 2.f * value_t::One(), 3.f * value_t::One()}};

   // All results in the vector are the same, so only check the first one
```

I.e. we should just leave it up to the constructor of `std::array` to do its thing, and just put an extra pair of parentheses in the client code where we want to initialize a vector from its elements. 🤔

krasznaa:
Note that unrelated to the templating issues I also had to make these changes:

```diff
diff --git a/benchmarks/common/include/benchmark/common/benchmark_vector.hpp b/benchmarks/common/include/benchmark/common/benchmark_vector.hpp
index 4000dda..2ddd055 100644
--- a/benchmarks/common/include/benchmark/common/benchmark_vector.hpp
+++ b/benchmarks/common/include/benchmark/common/benchmark_vector.hpp
@@ -71,7 +71,7 @@ struct vector_unaryOP_bm : public vector_bm<vector_t<scalar_t>> {
     const std::size_t n_warmup{this->m_cfg.n_warmup()};

     // Spin down before benchmark (Thread zero is counting the clock)
-    if (state.thread_index() == 0 and this->m_cfg.do_sleep()) {
+    if (state.thread_index() == 0 && this->m_cfg.do_sleep()) {
       std::this_thread::sleep_for(std::chrono::seconds(this->m_cfg.n_sleep()));
     }

@@ -113,7 +113,7 @@ struct vector_binaryOP_bm : public vector_bm<vector_t<scalar_t>> {
     const std::size_t n_warmup{this->m_cfg.n_warmup()};

     // Spin down before benchmark (Thread zero is counting the clock)
-    if ((state.thread_index() == 0) and this->m_cfg.do_sleep()) {
+    if ((state.thread_index() == 0) && this->m_cfg.do_sleep()) {
       std::this_thread::sleep_for(std::chrono::seconds(this->m_cfg.n_sleep()));
     }
```

MSVC just does not like the `and` keyword. 😦

storage/common/include/algebra/storage/vector.hpp
niermann999: Oh, I saw this post 2 seconds after trying for myself :) Hm, removing this constructor also removes much of the purpose of this class. The point is the call to `zero_fill` (which admittedly does not work right now, because of a Vc deprecation warning that kills the CI), which will correctly initialize a vector register that is larger than the number of passed arguments. I doubt the constructor of `std::array` can do that for us, nor should the client code in detray have to worry about this

niermann999:
> Note that unrelated to the templating issues I also had to make these changes:
> 
> ```diff
> diff --git a/benchmarks/common/include/benchmark/common/benchmark_vector.hpp b/benchmarks/common/include/benchmark/common/benchmark_vector.hpp
> index 4000dda..2ddd055 100644
> --- a/benchmarks/common/include/benchmark/common/benchmark_vector.hpp
> +++ b/benchmarks/common/include/benchmark/common/benchmark_vector.hpp
> @@ -71,7 +71,7 @@ struct vector_unaryOP_bm : public vector_bm<vector_t<scalar_t>> {
>      const std::size_t n_warmup{this->m_cfg.n_warmup()};
> 
>      // Spin down before benchmark (Thread zero is counting the clock)
> -    if (state.thread_index() == 0 and this->m_cfg.do_sleep()) {
> +    if (state.thread_index() == 0 && this->m_cfg.do_sleep()) {
>        std::this_thread::sleep_for(std::chrono::seconds(this->m_cfg.n_sleep()));
>      }
> 
> @@ -113,7 +113,7 @@ struct vector_binaryOP_bm : public vector_bm<vector_t<scalar_t>> {
>      const std::size_t n_warmup{this->m_cfg.n_warmup()};
> 
>      // Spin down before benchmark (Thread zero is counting the clock)
> -    if ((state.thread_index() == 0) and this->m_cfg.do_sleep()) {
> +    if ((state.thread_index() == 0) && this->m_cfg.do_sleep()) {
>        std::this_thread::sleep_for(std::chrono::seconds(this->m_cfg.n_sleep()));
>      }
> ```
> 
> MSVC just does not like the `and` keyword. 😦

Yes, I realized :) It is already gone from the PR #97 , but I have to put it here as well

storage/common/include/algebra/storage/vector.hpp
stephenswat: You could also add `template <typename... Values, std::enable_if_t<std::is_convertible_v<Values, scalar_t> && ..., bool> = true>`.

I'd add the forward either way, even if it's unlikely to matter for most scalar types.

niermann999: A `Vc::Vector` won't be convertible to a scalar (I hope :D), but maybe you are right, the issue with filling the vector register correctly only exists for the AoS case, where the arguments are indeed scalar/integer...

stephenswat: Uh yeah sorry I mean test if it's convertible to the relevant `Vc::Vector`. Why wouldn't that work?

niermann999: It could work, just pushed it :)

niermann999:
> P.S. I saw some "new" Eigen warnings from the CUDA build in my tests. Something to follow up on separately at one point. (Those may be only showing up on Windows, didn't test it rigorously.)

Btw, what were those warnings about? I started seeing some in detray, too, although it had previously stopped after we rolled back to version 34.0 (and now started again)

storage/common/include/algebra/storage/vector.hpp
niermann999: Looks good so far, thanks everyone!

storage/common/include/algebra/storage/vector.hpp
```diff
@@ -0,0 +1,246 @@
...
+  /// Construct from element values @param vals .
+  ///
+  /// In order to avoid uninitialized values, which deteriorate the performance
+  /// in explicitely vectorized code, the underlying data array is filled with
+  /// zeroes if too few arguments are given.
+  template <typename... Values,
+            std::enable_if_t<
+                std::conjunction_v<std::is_convertible<Values, value_type>...>,
+                bool> = true>
+  constexpr vector(Values &&... vals) : m_data{std::forward<Values>(vals)...} {
```
krasznaa, line 44: I'm fine with this type of SFINAE, this part is good on my part. In the morning I was also trying to make a similar thing work, but couldn't quite find the right incantation. So kudos to @stephenswat.

But there was one thing that Copilot very much wanted me to do while playing with this code: A check for `sizeof...(Values) <= N`. I think you could just add it with an `$$` after the `std::conjunction_v` expression.

krasznaa (commented):
With the Windows/MSVC issues gone, I'll try to find some time in the evening to have a look at the Mac issue. 🤔

In the future at one point the benchmarking code could also be simplified/unified a bit, but for this PR this setup should be fine as is...

benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+      LINK_LIBRARIES benchmark::benchmark algebra::bench_common
+                     algebra::bench_eigen algebra::eigen_eigen )
+   algebra_add_benchmark( eigen_vector
+      "eigen/eigen_vector.cpp"
+      LINK_LIBRARIES benchmark::benchmark algebra::bench_common
+                     algebra::bench_eigen algebra::eigen_eigen )
+endif()
+
+if( ALGEBRA_PLUGINS_INCLUDE_VC )
+   if( NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "AppleClang" )
```
krasznaa, line 56: I know that I opened a thread about this somewhere already, but couldn't find it. So here we go with a new one.

When I saw that Fastor was having problems on my M1 Mac, I was absolutely convinced that that would be the reason behind the CI's troubles with the benchmarks for the SoA code. But as it turns out, no. 😕

```
[zsh][Thranduil]:algebra > ./algebra-plugins/out/build/vc/bin/algebra_benchmark_vc_soa_getter
Algebra-Plugins 'getter' benchmark (Vc SoA)
-------------------------------------------

(single)
 -> running:	 160000 samples
 -> warmup: 	 16000 samples

(double)
 -> running:	 160000 samples
 -> warmup: 	 16000 samples

Unable to determine clock rate from sysctl: hw.cpufrequency: No such file or directory
2023-10-14T07:32:48+02:00
Running ./algebra-plugins/out/build/vc/bin/algebra_benchmark_vc_soa_getter
Run on (8 X 24.1209 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x2)
Load Average: 1.61, 1.61, 2.29
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
vector_GETTER_NAME_single/process_time/threads:8     216412 ns      1500595 ns          360
vector_GETTER_NAME_double/process_time/threads:8     717059 ns      5376955 ns          136
vector_GETTER_NAME_single/process_time/threads:8     218412 ns      1630665 ns          472
vector_GETTER_NAME_double/process_time/threads:8     807922 ns      6015875 ns          120
vector_GETTER_NAME_single/process_time/threads:8      19488 ns       146199 ns         5288
vector_GETTER_NAME_double/process_time/threads:8      19847 ns       147177 ns         5936
vector_GETTER_NAME_single/process_time/threads:8      20878 ns       148881 ns         4960
vector_GETTER_NAME_double/process_time/threads:8      21053 ns       152908 ns         5160
vector_GETTER_NAME_single/process_time/threads:8     275938 ns      2077637 ns          360
vector_GETTER_NAME_double/process_time/threads:8     394006 ns      2770471 ns          288
[zsh][Thranduil]:algebra > ./algebra-plugins/out/build/vc/bin/algebra_benchmark_vc_soa_vector
Algebra-Plugins 'vector' benchmark (Vc SoA)
-------------------------------------------

(single)
 -> running:	 160000 samples
 -> warmup: 	 16000 samples

(double)
 -> running:	 160000 samples
 -> warmup: 	 16000 samples

Unable to determine clock rate from sysctl: hw.cpufrequency: No such file or directory
2023-10-14T07:32:52+02:00
Running ./algebra-plugins/out/build/vc/bin/algebra_benchmark_vc_soa_vector
Run on (8 X 24.1208 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x2)
Load Average: 1.56, 1.60, 2.28
-----------------------------------------------------------------------------------------
Benchmark                                               Time             CPU   Iterations
-----------------------------------------------------------------------------------------
vector_add_single/process_time/threads:8            29176 ns       215771 ns         2440
vector_add_double/process_time/threads:8            33332 ns       249757 ns         2912
vector_sub_single/process_time/threads:8            28771 ns       210642 ns         3688
vector_sub_double/process_time/threads:8            33545 ns       246659 ns         3304
vector_dot_single/process_time/threads:8            27937 ns       200166 ns         3808
vector_dot_double/process_time/threads:8            33460 ns       247557 ns         2920
vector_cross_single/process_time/threads:8          32187 ns       241733 ns         3064
vector_cross_double/process_time/threads:8          36558 ns       275813 ns         2728
vector_normalize_single/process_time/threads:8      38788 ns       293758 ns         2664
vector_normalize_double/process_time/threads:8      40530 ns       298386 ns         2416
[zsh][Thranduil]:algebra >
```

My M1 chip seems to be just fine with these instructions. Much because Vc doesn't actually do anything on an ARM chip at the moment.

```
-- Building Vc as part of the Algebra Plugins project
-- Detected Compiler: Clang 15.0.0
CMake Warning at /Users/krasznaa/ATLAS/projects/algebra/build/_deps/vc-src/CMakeLists.txt:34 (message):
  No optimized implementation of the Vc types available for arm64
```

I'll revive my old MacBook for this, just to see if I can even reproduce the issues that you've been seeing in the CI... 🤔

storage/common/include/algebra/storage/vector.hpp
niermann999: Done

benchmarks/CMakeLists.txt
niermann999: It seemed to fail only in release mode, if I remember correctly, so only on specific instructions

krasznaa: Could you try re-enabling it in the CI, just so that I could see how it fails? I can't reproduce it on my 2014 MacBook Pro either.

The issue is either gone through some recent modification, or it's very specific to the CPU used in the GitHub runners. :confused:

stephenswat: Is the code being compiled for a native target by default? ARM chips have a lot more variety in what features they do and don't support which makes it extra important to compile for the right target.

stephenswat: Can you try printing `Vc_COMPILE_FLAGS`, `Vc_ARCHITECTURE_FLAGS` and `cat /proc/cpuinfo |grep flags`?

I know Vc does not support ARM, and I suspect that it is probably not setting any architecture flag at all, or it is setting some incompatible flag.

You could also try adding a CXXFLAGS="-march=native"` environment flag and let the compiler figure it out.

stephenswat: Or you could try `CXXFLAGS="-march=armv8-a"` and see if that works.

niermann999: I don't know if this is helps, but during configuration of Vc the XCode CI prints:
```
-- Detected CPU: ivy-bridge
-- Performing Test check_cxx_compiler_flag__march_ivybridge
-- Performing Test check_cxx_compiler_flag__march_ivybridge - Success
```

stephenswat: Oh I thought this was about ARM. :laughing: Never mind.

krasznaa: Unfortunately I'm running out of ideas of how to debug this further. 😦

The CI nodes used macOS 12.7 and Xcode 14. On my old MacBook the latest software that I can install is macOS 11.7 and Xcode 13. 😕 And as the Vc configuration reports, my laptop has a Haswell CPU. While the CI runs on Ivy Bridge Xeon CPUs as it seems.

As far as I can see, the CI uses the same optimization settings for the build

```
CompileC /Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/Objects-normal/x86_64/vc_soa.o /Users/runner/work/algebra-plugins/algebra-plugins/tests/vc_soa/vc_soa.cpp normal x86_64 c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'algebra_test_vc_soa' from project 'algebra-plugins')
    cd /Users/runner/work/algebra-plugins/algebra-plugins
    /Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x c++ -target x86_64-apple-macos12.7 -fmessage-length\=0 -fdiagnostics-show-note-include-stack -fmacro-backtrace-limit\=0 -Wno-trigraphs -fpascal-strings -O3 -Wno-missing-field-initializers -Wno-missing-prototypes -Wno-return-type -Wno-non-virtual-dtor -Wno-overloaded-virtual -Wno-exit-time-destructors -Wno-missing-braces -Wparentheses -Wswitch -Wno-unused-function -Wno-unused-label -Wno-unused-parameter -Wno-unused-variable -Wunused-value -Wno-empty-body -Wno-uninitialized -Wno-unknown-pragmas -Wno-shadow -Wno-four-char-constants -Wno-conversion -Wno-constant-conversion -Wno-int-conversion -Wno-bool-conversion -Wno-enum-conversion -Wno-float-conversion -Wno-non-literal-null-conversion -Wno-objc-literal-conversion -Wshorten-64-to-32 -Wno-newline-eof -Wno-c++11-extensions -Wno-implicit-fallthrough -DCMAKE_INTDIR\=\"Release\" -isysroot /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk -fasm-blocks -fstrict-aliasing -Wdeprecated-declarations -Winvalid-offsetof -Wno-sign-conversion -Wno-infinite-recursion -Wno-move -Wno-comma -Wno-block-capture-autoreleasing -Wno-strict-prototypes -Wno-range-loop-analysis -Wno-semicolon-before-method-body -I/Users/runner/work/algebra-plugins/algebra-plugins/build/bin/Release/include -I/Users/runner/work/algebra-plugins/algebra-plugins/storage/vc_soa/include -I/Users/runner/work/algebra-plugins/algebra-plugins/common/include -I/Users/runner/work/algebra-plugins/algebra-plugins/storage/common/include -I/Users/runner/work/algebra-plugins/algebra-plugins/frontend/vc_soa/include -I/Users/runner/work/algebra-plugins/algebra-plugins/math/cmath/include -I/Users/runner/work/algebra-plugins/algebra-plugins/math/common/include -I/Users/runner/work/algebra-plugins/algebra-plugins/math/vc_soa/include -isystem /Users/runner/work/algebra-plugins/algebra-plugins/build/_deps/googletest-src/googletest/include -isystem /Users/runner/work/algebra-plugins/algebra-plugins/build/_deps/googletest-src/googletest -isystem /Users/runner/work/algebra-plugins/algebra-plugins/build/_deps/vc-src -I/Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/DerivedSources-normal/x86_64 -I/Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/DerivedSources/x86_64 -I/Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/DerivedSources -F/Users/runner/work/algebra-plugins/algebra-plugins/build/bin/Release -Wall -Wextra -Wshadow -Wunused-local-typedefs -pedantic -Werror -DNDEBUG -std\=c++17 -Wno-local-type-template-args -Wno-unnamed-type-template-args -ffp-contract\=fast -march\=ivybridge -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mavx -mno-sse4a -mno-fma -mno-bmi2 -mno-avx2 -mno-xop -mno-fma4 -mno-avx512f -mno-avx512vl -mno-avx512pf -mno-avx512er -mno-avx512cd -mno-avx512dq -mno-avx512bw -mno-avx512ifma -mno-avx512vbmi -MMD -MT dependencies -MF /Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/Objects-normal/x86_64/vc_soa.d --serialize-diagnostics /Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/Objects-normal/x86_64/vc_soa.dia -c /Users/runner/work/algebra-plugins/algebra-plugins/tests/vc_soa/vc_soa.cpp -o /Users/runner/work/algebra-plugins/algebra-plugins/build/build/algebra_test_vc_soa.build/Release/Objects-normal/x86_64/vc_soa.o
```

as my laptop does:

```
[zsh][tauriel]:algebra > more build/tests/CMakeFiles/algebra_test_vc_soa.dir/flags.make 
# CMAKE generated file: DO NOT EDIT!
# Generated by "Unix Makefiles" Generator, CMake Version 3.27

# compile CXX with /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
CXX_DEFINES = 

CXX_INCLUDES = -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/storage/vc_soa/include -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/common/include -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/storage/common/include -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/frontend/vc_soa/include -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/math/cmath/include -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/math/common/include -I/Users/krasznaa/ATLAS/projects/algebra/algebra-plugins/math/vc_soa/include -isystem /Users/krasznaa/ATLAS/projects/algebra/build/_deps/googletest-src/googletest/include -isystem /Users/krasznaa/ATLAS/projects/algebra/build/_deps/googletest-src/googletest -isystem /Users/krasznaa/ATLAS/projects/algebra/build/_deps/vc-src

CXX_FLAGS =  -Wall -Wextra -Wshadow -Wunused-local-typedefs -pedantic -Werror -O3 -DNDEBUG -std=c++17 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk -mmacosx-version-min=11.7 -Wno-local-type-template-args -Wno-unnamed-type-template-args -ffp-contract=fast -march=haswell -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mavx -mfma -mbmi2 -mavx2 -mno-sse4a -mno-xop -mno-fma4 -mno-avx512f -mno-avx512vl -mno-avx512pf -mno-avx512er -mno-avx512cd -mno-avx512dq -mno-avx512bw -mno-avx512ifma -mno-avx512vbmi
```

As a final thing I downloaded the binaries produced by the CI, from: https://github.com/krasznaa/algebra-plugins/actions/runs/6535189458 But unfortunately, even after I told macOS over and over again for every binary in that ZIP file that it's allowed to use it, I still get:

```
[zsh][tauriel]:Release > DYLD_LIBRARY_PATH=/Users/krasznaa/ATLAS/projects/algebra/ci/lib/Release ./algebra_test_vc_soa
dyld: Symbol not found: __ZNKSt3__115basic_stringbufIcNS_11char_traitsIcEENS_9allocatorIcEEE3strEv
  Referenced from: /Users/krasznaa/ATLAS/projects/algebra/ci/lib/Release/libgtest.1.14.0.dylib (which was built for Mac OS X 12.7)
  Expected in: /usr/lib/libc++.1.dylib

zsh: abort      DYLD_LIBRARY_PATH=/Users/krasznaa/ATLAS/projects/algebra/ci/lib/Release 
[zsh][tauriel]:Release > 
```

Since macOS is just not forward compatible. 😦

The flags used in both builds seem correct to me. 🤔 They both go up to AVX instructions. And according to Wikipedia, Icy Bridge is supposed to be able to do that. 😕 

So at this point I'll just do one more thing, I'll turn off the Vc compiler optimizations for a test build, and see if that works. If it does, we should just turn off those platform specific optimizations on macOS, and not the tests as a whole...

krasznaa: Huhh... There is a twist. 😕 After I switched off the platform specific optimizations for this project's code with https://github.com/krasznaa/algebra-plugins/commit/0f7276e790c6468826b829eb5c54fce577883af8, the test still dies with an invalid instruction. 🤔

https://github.com/krasznaa/algebra-plugins/actions/runs/6535685632/job/17745684927

So it seems that it's something with Vc itself. 🤔 It seems that it's unable to choose the correct implementation of one of its optimized functions in that build. Interesting...

I now blame macOS 12.7 for not choosing the correct functions at runtime. One more thing that I'll try, after that I'll be out of ideas...

stephenswat: Running the code through LLDB produces no error, so I strongly suspect that this is not actually related to compiler flags, but rather that we are producing some undefined behaviour somewhere. I'll try again with ASLR enabled to see if that re-introduces the bug.

krasznaa: You mean you ran the debugger in the CI?

Because if not, then I'm very surprised that we're still not on the same page with this. 😕 The binary produced by the CI clearly contains an instruction that the build node's Ivy Bridge cpu cannot handle.

We've seen a similar issue in the offline a while ago. That's why "invalid instruction" seems pretty clear to me. I don't see signs of undefined behviour anywhere here...

stephenswat: Yes, I'm running LLDB in the CI. :slightly_smiling_face:

stephenswat: Okay, so here's the fun stuff. LLDB in the CI returns the following:

```
Running main() from /Users/runner/work/algebra-plugins/algebra-plugins/build/_deps/googletest-src/googletest/src/gtest_main.cc
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from test_vc_host
[ RUN      ] test_vc_host.vc_soa_vector
[       OK ] test_vc_host.vc_soa_vector (0 ms)
[ RUN      ] test_vc_host.vc_soa_getter
Process 4114 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
    frame #0: 0x00000001010da1b2 algebra_test_vc_soa`Vc_1::Vector<float, Vc_1::VectorAbi::Avx> Vc_1::Common::Trigonometric<Vc_1::ImplementationT<6u> >::atan2<Vc_1::Vector<float, Vc_1::VectorAbi::Avx> >(Vc_1::Vector<float, Vc_1::VectorAbi::Avx> const&, Vc_1::Vector<float, Vc_1::VectorAbi::Avx> const&) + 242
algebra_test_vc_soa`Vc_1::Common::Trigonometric<Vc_1::ImplementationT<6u> >::atan2<Vc_1::Vector<float, Vc_1::VectorAbi::Avx> >:
->  0x1010da1b2 <+242>: vfmsub231ps %ymm14, %ymm13, %ymm0     ; ymm0 = (ymm13 * ymm14) - ymm0
    0x1010da1b7 <+247>: vbroadcastss 0x38(%rax), %ymm14
    0x1010da1bd <+253>: vfmadd231ps %ymm0, %ymm13, %ymm14     ; ymm14 = (ymm13 * ymm0) + ymm14
    0x1010da1c2 <+258>: vbroadcastss 0x3c(%rax), %ymm0
Target 0: (algebra_test_vc_soa) stopped.
Process 4114 launched: '/Users/runner/work/algebra-plugins/algebra-plugins/build/bin/Release/algebra_test_vc_soa' (x86_64)
```

Interesting thing to note here:

1. The process is failing on `vfmsub231ps` which is a FMA3 instruction which Ivy Bridge does not support, so it is not a jump to an invalid instruction as I suspected.
2. The implementation identifier is 6, which [translates to AVX](https://github.com/VcDevel/Vc/blob/93d646eafb6714164eb28eb9571cbca6985572a0/Vc/global.h#L496), which is correct.
3. The `0x20000` bit [which indicates FMA3](https://github.com/VcDevel/Vc/blob/93d646eafb6714164eb28eb9571cbca6985572a0/Vc/global.h#L526) is _not_ set, which is correct.
4. Vc is still emitting a `vfmsub231ps` instruction in its implementation of `atan2` for AVX architectures without FMA.
5. It seems likely, but not certain, that [this subtraction](https://github.com/VcDevel/Vc/blob/1.4.3/src/trigonometric.cpp#L556) is the cause of the incorrect code emission.
6. The Vc code base does not appear to contain any mentions of `fmsub` intrinsics, so I would assume the code is automatically generated.

Thus:

1. Vc is jumping into the correct functions at runtime.
2. The assembly code for `libVc.a` is being generated incorrectly.

Probably more interesting things to be found but not today.

stephenswat: Well, things turned out to be too much fun to leave until tomorrow. :smile: I haven't tested it, but I am quite certain I understand what is going on; it's a case of the compiler being smarter than the build system compiler.

First, a primer on how Vc is built. Vc produces a fat binary object `libVc.a` which is designed to run anywhere, and to run everywhere efficiently. Because the machine on which the code will be run is not known at compile time, `libVc.a` must contain multiple versions of all of its functions, for SSE2, SSE4.2, AVX, AVX2, etc.

Because the infrastructure for generating multiple versions of functions in the same translation unit is relatively poorly defined in most compilers (although some can do it), Vc achieves fat binary production by producing a large number of small binaries and linking them together. The trigonometry part of Vc is compiled into the following objects:

* `trigonometric_AVX+XOP+FMA.o`
* `trigonometric_AVX+XOP+FMA4.o`
* `trigonometric_SSE4_1.o`
* `trigonometric_SSE3.o`
* `trigonometric_SSSE3.o`
* `trigonometric_SSE+XOP+FMA.o`
* `trigonometric_AVX+FMA.o`
* `trigonometric_AVX2+FMA+BMI2.o`
* `trigonometric_AVX.o`
* `trigonometric_SSE+XOP+FMA4.o`
* `trigonometric_SSE2.o`

All of this is well and good. Let's inspect the `atan2` function (which we know to be problematic) in the `trigonometric_AVX.o` file, which is the one that will be used on the Github runners:

```
00000000000020f0 <Vc_1::Vector<float, Vc_1::VectorAbi::Avx> Vc_1::Common::Trigonometric<Vc_1::ImplementationT<6u>>::atan2<Vc_1::Vector<float, Vc_1::VectorAbi::Avx>>(Vc_1::Vector<float, Vc_1::VectorAbi::Avx> const
&, Vc_1::Vector<float, Vc_1::VectorAbi::Avx> const&)>:
[some assembly redacted for brevity]
    21d6: c4 62 7d 18 70 30             vbroadcastss    48(%rax), %ymm14
    21dc: c4 41 14 59 f6                vmulps  %ymm14, %ymm13, %ymm14
    21e1: c4 e2 7d 18 60 34             vbroadcastss    52(%rax), %ymm4
    21e7: c5 8c 5c e4                   vsubps  %ymm4, %ymm14, %ymm4
    21eb: c5 94 59 e4                   vmulps  %ymm4, %ymm13, %ymm4
    21ef: c4 62 7d 18 70 38             vbroadcastss    56(%rax), %ymm14
    21f5: c5 8c 58 e4                   vaddps  %ymm4, %ymm14, %ymm4
    21f9: c5 94 59 e4                   vmulps  %ymm4, %ymm13, %ymm4
    21fd: c4 62 7d 18 70 3c             vbroadcastss    60(%rax), %ymm14
    2203: c4 c1 5c 5c e6                vsubps  %ymm14, %ymm4, %ymm4
    2208: c5 94 59 e4                   vmulps  %ymm4, %ymm13, %ymm4
```

Looks good! This is all AVX code which does not use FMA. All is well, and this would run perfectly on the CI runner. Note the instructions that are being run: vector subtractions (`vsubps`) and additions (`vaddps`) followed by multiplications (`vmulps`). Let's find the same code in the fat binary `libVc.a`:

```
0000000000002460 <Vc_1::Vector<float, Vc_1::VectorAbi::Avx> Vc_1::Common::Trigonometric<Vc_1::ImplementationT<7u>>::atan2<Vc_1::Vector<float, Vc_1::VectorAbi::Avx>>(Vc_1::Vector<float, Vc_1::VectorAbi::Avx> const
&, Vc_1::Vector<float, Vc_1::VectorAbi::Avx> const&)>:
[some assembly redacted for brevity]
    2547: c4 62 7d 18 70 30             vbroadcastss    48(%rax), %ymm14
    254d: c4 e2 7d 18 40 34             vbroadcastss    52(%rax), %ymm0
    2553: c4 c2 15 ba c6                vfmsub231ps     %ymm14, %ymm13, %ymm0 ## ymm0 = (ymm13 * ymm14) - ymm0
    2558: c4 62 7d 18 70 38             vbroadcastss    56(%rax), %ymm14
    255e: c4 62 15 b8 f0                vfmadd231ps     %ymm0, %ymm13, %ymm14 ## ymm14 = (ymm13 * ymm0) + ymm14
    2563: c4 e2 7d 18 40 3c             vbroadcastss    60(%rax), %ymm0
    2569: c4 c2 15 ba c6                vfmsub231ps     %ymm14, %ymm13, %ymm0 ## ymm0 = (ymm13 * ymm14) - ymm0
    256e: c5 94 59 c0                   vmulps  %ymm0, %ymm13, %ymm0
```

Two things to notice here. First, the code has _changed_ during linking. Second, the code is _equivalent_ to the original code. How did this happen?

The answer is link-time optimization. Vc takes great care to ensure that the individual translation units are compiled with the right settings, but the linker is _not_ told to respect any of this; the linker is free to optimize at it pleases. In this case, the linker has performed link-time peephole optimization or, more specifically, link-time floating point contraction, which is the contraction of additions (or subtractions) and multiplications into fused multiply-addition operations!

How cool is that? :smile: Presumably the Mach-O binary format used by XCode does not store information about compilation targets to ensure that this doesn't happen, but I don't know.

stephenswat: Confirmed, the fix was extremely simple:

First, I had to patch Vc to disable link-time floating point contraction when using XCode, which I did in https://github.com/stephenswat/Vc/commit/d56339236d9014c6e78e673df4450c977b04a078

Second, I updated the version of Vc used to my patched version in https://github.com/stephenswat/algebra-plugins/commit/e0c40e168588ec5f41740e9af367338802cfdda3

This happily passes the CI: https://github.com/stephenswat/algebra-plugins/actions/runs/6540226183/job/17759772680

niermann999: Nvm, I just found it :D

paulgessinger: @stephenswat dude.... 😮

niermann999: As it seems that this is a bit stuck with VC, should we add the flag here for corresponding version of xcode?

tests/vc_soa/vc_soa.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <iostream>
+
+using namespace algebra;
+
+using scalar_t = float;
+
+constexpr float tol{1e-6f};
+
+/// This test the vector functions on an SoA (Vc::Vector) based vector
+TEST(test_vc_host, vc_soa_vector) {
```
beomki-yeo, line 24: Is this test really for SoA (multiple vectors)? It looks like it is testing operations between two vectors

niermann999: Each object of type ```vc_soa::vector3<scalar_t>;``` contains a number of vectors equal to the simd width, if that is what you mean.

niermann999: So, ```a```, for example will look like this (float, avx2):
```
[1,1, 1, 1, 1, 1, 1, 1]
[2, 2, 2, 2, 2, 2, 2, 2]
[3, 3, 3, 3, 3, 3, 3, 3]
```
In an actual application, the numbers should not all be the same of course, but it was easier to write the test this way. There is, for example, no parametrized constructor for ```Vc::Vector```, if I remember correctly, because the number of elements it can hold depends on the floating point precision and the hardware capabilities

beomki-yeo: Yes that's what I meant. I was just ignorant.
To make the tests more reliable, I would put different vectors for each simd element. But that's your choice

beomki-yeo: Our comments get crossed.
 The number of SIMD elements cannot be accessed in run time?

niermann999: No, it depends on the floating point precision and on the particular CPU, because the number of bits in the vector registers is fixed. So, if I want to fill the a single simd element with different values, I have to make that dependent on these conditions as well

niermann999: Ah, wait accessed? It can, and has to be accessed at compile time, actually. I just cannot write
````
Vc::Vector<scalar> a{1, 2, 3, 4, 5, 6, 7, 8};
```
or rely on the fact that ```a[6]``` exists.

niermann999: I think it would be easy to do with a ```Vc``` inbuilt, though. I just did something similar in detray... Then we would be able to check that ```Vc``` implements its operations correctly, I guess

beomki-yeo: Maybe it can be accessed with a predefined variable:
https://vcdevel.github.io/Vc-1.4/vcmacros.html#vc_size

But I don't know anything about it so you need to judge it

beomki-yeo: Maybe you meant that it is difficult to decide whether to use double and float size, so yeah just forget about it if it is a hassle

beomki-yeo: Thought there is a  `Vector::size()` that seems to be accessible in runtime

niermann999: Yes, this is even accessible at compile-time. But I still need a for loop or a lambda and a prescription on how to set the entries for every element of the soa vector and also for the results at every step to compare them against, right? Because, then I cannot compare against ```value_t(1.f)``` etc. anymore. I am not saying it is impossible, I am just not sure I have time for that now

niermann999: What I meant with the Vc inbuilt is ```Vc::Vector<scalar>::IndexesFromZero()``` which will give me reasonable starting vectors at least

beomki-yeo: You don't have to do that if the time matters. 
Just FYI, you can put array with any pattern. For example, for n-th vector of `a` is set to [n, n+1, n+2]  and n-th vector of `b` to [n,n-1,n-2] and you can expect the output of addition, cross product, etc.

niermann999: Yes, this is pretty much what ```IndexesFromZero()``` allows you to do as a one-liner. I just don't have the time left to actually do it...

benchmarks/CMakeLists.txt
krasznaa: Since none of us were active on providing a better fix for this, let's just keep the benchmark turned off on macOS for now. We can fix this in a better way in the future, based on what we figured out last year...

storage/vc/CMakeLists.txt
```diff
@@ -9,6 +9,6 @@ algebra_add_library( algebra_vc_storage vc_storage
    "include/algebra/storage/vc.hpp"
    "include/algebra/storage/impl/vc_array4_wrapper.hpp" )
 target_link_libraries( algebra_vc_storage
-   INTERFACE algebra::common Vc::Vc )
+   INTERFACE algebra::common algebra::common_storage Vc::Vc )
```
krasznaa, line 12: :confused: I don't see any of the code in `algebra::vc_storage` changing in this PR in the end. I believe the change in this line should be undone. :thinking:

tests/CMakeLists.txt
```diff
@@ -41,7 +41,14 @@ if( ALGEBRA_PLUGINS_INCLUDE_VC )
    algebra_add_test( vc
       "vc/vc_cmath.cpp" "vc/vc_vc.cpp"
       LINK_LIBRARIES GTest::gtest_main algebra::tests_common
-                     algebra::vc_cmath algebra::vc_vc )
+                      algebra::vc_storage algebra::vc_cmath algebra::vc_vc )
```
krasznaa, line 44: This line should not need a change. Just linking against the "frontent" libraries should be enough. :thinking:

tests/CMakeLists.txt
```diff
@@ -41,7 +41,14 @@ if( ALGEBRA_PLUGINS_INCLUDE_VC )
...
+
+   if( NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "AppleClang" )
+      algebra_add_test( vc_soa
+         "vc_soa/vc_soa.cpp"
+         LINK_LIBRARIES GTest::gtest_main algebra::vc_soa_storage 
+                        algebra::vc_soa )
```
krasznaa, line 50: It should be enough to just link against `algebra::vc_soa` here as well. :thinking: You should not link against the underlying `algebra::vc_soa_storage` library explicitly as well.

krasznaa (changes requested):
Code-wise I don't have objections. Just some smaller cmake updates, and I'm happy for this to go in. :smile:

tests/vc_soa/vc_soa.cpp
beomki-yeo: Ah yes

tests/CMakeLists.txt
niermann999: Done

tests/CMakeLists.txt
niermann999: Done

krasznaa (approved):
Everything should be resolved now... :thinking:

PR #96, opened 2023-03-24
CMake Export Fix, main branch (2023.03.24.)

### LINKED ISSUE #36: Algebra plugins configurations not working as intended
So here's an interesting little CMake problem. I'm trying to compile detray against a system installation of the algebra plugins, which is giving me the following configuration-time error:

```
CMake Error at /lib/cmake/Vc/VcConfig.cmake:11 (message):
  File or directory
  /lib/cmake/algebra-plugins-0.2
  referenced by variable algebra_plugins_CMAKE_DIR does not exist !
Call Stack (most recent call first):
  /lib/cmake/algebra-plugins-0.2/algebra-plugins-config.cmake:51 (set_and_check)
  CMakeLists.txt:67 (find_package)
```

I'm suspecting that this is a bug in the algebra plugins build system, specifically in `/cmake/algebra-plugins-config.cmake.in`. I don't really understand why the Vc configuration is trying to look for the algebra plugins here, that seems like the opposite of what should be happening.

The buggy line in question (`/lib/cmake/algebra-plugins-0.2/algebra-plugins-config.cmake:51`) is the following:

```cmake
set_and_check( algebra_plugins_CMAKE_DIR "${PACKAGE_PREFIX_DIR}/lib/cmake/algebra-plugins-0.2" )
```

Maybe @krasznaa has any ideas on what might be going on?

krasznaa:
Moved the use of "magic variables" before the first use of find_dependency to fix #36. (This of course also very much relates to https://github.com/acts-project/traccc/pull/364.)

The way that the variables provided by [configure_package_config_file(...)](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:configure_package_config_file) work is that a "`PACKAGE_PREFIX_DIR`" variable is created behind the scenes, and all other variables are declared relative to this "hidden" variable.

The problem is that if a package "found" by [find_dependency(...)](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) uses the same CMake mechanism in its own exported CMake configuration (like [Vc](https://github.com/VcDevel/Vc) does for instance), then this "hidden" `PACKAGE_PREFIX_DIR` variable gets re-defined after the `find_dependency(...)` calls. And all of a sudden the rest of the lines that try to set paths using the mechanism provided by `configure_package_config_file(...)` become broken in a fairly confusing way.

This is something that will need to be fixed in the rest of our projects as well, Algebra Plugins is just the first in line to be fixed...

### DIFF
--- a/cmake/algebra-plugins-config.cmake.in
+++ b/cmake/algebra-plugins-config.cmake.in
@@ -1,6 +1,6 @@
 # Algebra plugins library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2022 CERN for the benefit of the ACTS project
+# (c) 2021-2023 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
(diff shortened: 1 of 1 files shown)


stephenswat (approved):
Appears functional after brief test. :+1:

PR #97, opened 2023-03-26
feat: add a Vc SoA implementation for transform3

niermann999:
This PR is based on the previous development for an SoA Vc-based algebra plugin and adds the ```transform3``` implementation, including a test and benchmarks. Like the current ```vc_vc``` plugin, it uses the ```vector3``` type as column vectors in the 4x4 matrix type that is used by the ```transform3```. Elements that are known to be equal to zero or one are optimized away in the inversion and determinant calculations.


niermann999:
Based on #95

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
```diff
@@ -0,0 +1,386 @@
...
+using algebra::storage::operator/;
+using algebra::storage::operator-;
+using algebra::storage::operator+;
+
+namespace internal {
+
+/// 4x4 matrix type used by @c algebra::vc_soa::math::transform3 that has simd
+/// vectors as matrix elements
+template <typename scalar_t, template <typename, std::size_t> class array_t>
+struct matrix44 {
```
beomki-yeo, line 40: Why don't you use the stroage matrix type of `vc_soa` algebra?

beomki-yeo:
And I suggest separating a PR for benchmark

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  /// The determinant of a 4x4 matrix
+  ///
+  /// @param m is the matrix
+  ///
+  /// @return a sacalar determinant - no checking done
+  ALGEBRA_HOST_DEVICE
+  inline constexpr value_type determinant(const matrix44 &m) const {
+    return m.z[0] * m.y[1] * m.x[2] + m.y[0] * m.z[1] * m.x[2] +
```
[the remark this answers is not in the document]
niermann999: I think I took the implementation from the other Vc plugin and then removed the terms that are always zero or one, because the transform3 is in homogeneous coordinates (```x[3]```, ```y[3]```, ```z[3]``` are zero and ```t[3]``` is one)

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
niermann999: I am ```using vector_type = storage::vector<3, Vc::Vector<scalar_t>, array_t>;``` for the ```matrix44``` columns. This way I can use e.g. the cross product on them and easily build a matrix from ```vector3``` types. If I use ```storage::vector<ROWS * COLS>```, I just have a ```Vc::Vector<T>``` type for every matrix element, for which I have no linear algebra implementation. 
So say I want to implement ```rotate```: Right now that is simple but with ```storage::matrix``` I would have to write: {m[0][0] * v[0] + m[0][1] * v[1] + m[0][2] * v[2], m[1][0] * v[0] + m[1][1] * v[1] + m[1][2] * v[2], ...}```

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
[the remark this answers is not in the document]
beomki-yeo: I think the determinant of transform3 is:
       [x0 y0 z0 t0]          [ x0 y0 z0 ]
det  [x1 y1 z1 t1] = det [ x1 y1 z1 ] = x0 [ y1 * z2 - z1 y2 ] - y0 [ x1 * z2 - z1 * x2 ] + z0 [ x1 * y2 - y1 * x2 ]
       [x2 y2 z2 t2]          [ x2 y2 z2 ]
       [0  0   0    1]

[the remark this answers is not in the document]
niermann999: Ah, I see what happened, I forgot to copy paste the minus sign of the first term, because its written in the line above. Thanks!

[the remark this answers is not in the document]
niermann999: Odd that the unittest passes...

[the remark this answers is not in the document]
beomki-yeo: x_2 is set to zero in the soa test.

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
beomki-yeo: I suppose It is because there is no matrix operation for `vc_soa::matrix_type`
Is it possible to make one for `vc_soa::matrix_type` or it will just degrade the performance of SoA?

niermann999: Mainly, a matrix implementation would be a lot of extra work at this point which is not needed so far. If there is a usecase in traccc we can add it later

The last row of that matrix would eat a bit of memory: E.g. on my laptop [[0,0,0,0,0,0,0,0], [0,0,0,0,0,0,0,0], [0,0,0,0,0,0,0,0], [1,1,1,1,1,1,1,1]] for both the matrix and the inverse matrix per surface. But we could make it contiguous in memory in a rowwise layout, so it should not be that bad in terms of memory access (it could be that this block will then simply not be loaded since the transform3 implementation does not access it). 

However, I don't think it is possible for the AoS Vc implementation to do that (we might be able to share ```matrix44``` between the two plugins later, but I am not sure yet)

beomki-yeo: I guess `vc_soa` is being implemented for vectorization, isn't it? If so, it should support matrix operation as well as vector operation (Matrix operation is a thing that you might want to optimize the most...)

> The last row of that matrix would eat a bit of memory: E.g. on my laptop [[0,0,0,0,0,0,0,0], [0,0,0,0,0,0,0,0], [0,0,0,0,0,0,0,0], [1,1,1,1,1,1,1,1]] 

Is it from `matrix44` or Is it a thing that going to happen with `vc_soa::matrix_type`?

> I don't think it is possible for the AoS Vc implementation

Is `AoS Vc implementation` vc_vc frontend we have? We can forget about this frontend as it is not what we want from vc plugin

niermann999: That would happen with ```vc_soa::matrix_type``` if we don't make it an asymmetric matrix (which would be OK, I guess, since in SoA the memory alignment is independent from the matrix dimension). 
Yes, it is implemented for vectorization, but in detray at least I did not find a good spot for SoA matrices, yet? Even putting the static geometry into SoA layout is not easy...
I would rather go with an AoS vectorization scheme (```vc``` backend) for the propagation in detray (at least until we come up with a good idea so handle the track state divergence), so next step would be fixing the ```vc``` plugin and then adding a matrix implemenation to it. Why is the ```vc_vc``` frontend not what we want?

beomki-yeo: > in detray at least I did not find a good spot for SoA matrices, yet?

Using SoA matrix for covariance transport would be as interesting as usiing SoA vector for intersector

> Why is the vc_vc frontend not what we want?

I don't know its future form, but it is not using any actual `vc` function currently. Can it be more interesting in the future? Vc can be efficient for both AoS and SoA? My understanding on Vc plugin is not great...

niermann999: Yes, AoS and SoA are just memory layouts, while Vc is a library for explicit vectorization. Vectorization usually works better in an SoA layout, because you can fill the vector registers with more relevant data, but the old Vc plugin can vectorize the data in AoS layout and also achieve some speedup (especially 8x8 matrices with avx2).

The old Vc plugin implementation is fine actually, it was just incorrectly adapted to the cmath matrix actor, which broke it. But if we remove that matrix actor and write one that uses ```Vc::SimdArray``` internally, then we can use it for AoS vectorization. However, as far as I know, ```Vc::SimdArray``` is not made for linear algebra vectors that are much larger than the underlying vector register capacity, so for a generalized N-dimensional implementation, this could become a bit complicated.

Due to the track state divergence, I don't think an SoA memory layout makes sense to use in the covariance transport until we find a way to implement it such that we can pretty much guarantee that we can sort the tracks into chunks of 8, 16 , 32 etc. after seeding and then always access these tracks together in memory and have them follow the same algorithmic code path as well (or did you find a way in the CKF maybe?). Or alternatively, do you know of a way to use intra-track parallelization with SoA in the covariance transport (I thought the covariance matrices of a single track are strictly dependent on each other)?

beomki-yeo: >  (or did you find a way in the CKF maybe?).

Is that correct that you are using vc_soa to calculate the intersections of multiple candidates efficiently during navigator initialization? 
(Actually I don't understand what is the  difference between original intersector and SoA intersector you introduced in acts-project/detray#555)

If so, we can also do the KF for multiple measurements during CKF. I am not sure if you have looked into the results of [mkFit](https://iris-hep.org/projects/mkfit.html) of CMS collaboration, but track building part is the only place they achieved the speedup with vectorized matrix.

niermann999: > Is that correct that you are using vc_soa to calculate the intersections of multiple candidates efficiently

Yes, that is the plan, but the navigator has not been adapted to use these intersectors, yet. The CPU will be capable of intersecting multiple surfaces for a single track in parallel. But which surfaces can be intersected in parallel is pretty much a fixed thing.

> If so, we can also do the KF for multiple measurements during CKF. I am not sure if you have looked into the results of [mkFit](https://iris-hep.org/projects/mkfit.html) of CMS collaboration, but track building part is the only place they achieved the speedup with vectorized matrix.

Cool, I did not know that! That is definetly something we should consider then, can we go over their paper on Monday?

beomki-yeo: yes we can I guess. Andi must know about mkFit well

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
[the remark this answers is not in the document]
niermann999: Added a minus sign

niermann999:
> And I suggest separating a PR for benchmark

The benchmarks are already part of #95 which this one is based on

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
```diff
@@ -0,0 +1,386 @@
...
+    return ((x == rhs.x) && (y == rhs.y) && (z == rhs.z) && (t == rhs.t));
+  }
+
+  /// Data variables
+  vector_type x, y, z, t;
+
+};  // struct matrix44
+
+/// Functor used to access elements of matrix44
+template <typename scalar_t, template <typename, std::size_t> class array_t>
```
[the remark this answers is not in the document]
niermann999: I set it to ```std::array``` directly now

math/vc_soa/include/algebra/math/impl/vc_soa_transform3.hpp
```diff
@@ -0,0 +1,386 @@
...
+      case 0u:
+        return m.x[row];
+      case 1u:
+        return m.y[row];
+      case 2u:
+        return m.z[row];
+      case 3u:
+        return m.t[row];
+      default:
+        return m.x[0];
```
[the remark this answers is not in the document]
krasznaa: :thinking: Not exactly sure myself. I was hoping that you may know. :stuck_out_tongue:

This is not an actual worry of mine though. I was just bringing this up as something that may be possible to introduce a clear error for. Since the assertions are already in place to defend a debug build, we can just chalk it up as a thing that we may want to do in a more clever way later on.

krasznaa (approved):
Yeah, let's go forward with it...

PR #98, opened 2023-03-29
VecMem Update, main branch (2023.03.29.)

krasznaa:
Updated the project to use [vecmem-0.24.0](https://github.com/acts-project/vecmem/releases/tag/v0.24.0).

A bit surprisingly none of the tests in this project use vector buffers, so the code didn't actually need to be updated...


beomki-yeo (approved):
:+1:

krasznaa:
It was only a technical issue, but we should still wait for the CI to succeed before we would merge anything...

PR #99, opened 2023-03-29
Benchmarks Fix, main branch (2023.03.29.)

krasznaa:
Removed the reference to the non-existent `benchmarks/` directory. This was introduced in #69, which I just really don't see the point in now that I look back. :frowning: (One would only add Google Benchmark in the PR in which it is first needed. And touching all those files without a need is just pretty bad... :frowning:)

At the same time enabled the "build of benchmarks" in the CI. In case we ever add some to the project...


niermann999 (approved):

krasznaa:
That does paint it in a slightly different light, but let's just add back the `add_subdirectory(...)` call in #95 then. Once there is actually a subdirectory there. :wink:

PR #100, opened 2023-03-29
Public Header Tests, main branch (2023.03.29.)

krasznaa:
This is a sibling of https://github.com/acts-project/vecmem/pull/222.

I added a function called `algebra_test_public_headers(...)` that can be used to make sure that:
  - public header files include everything that they need to be usable;
  - the CMake library targets are set up correctly to make the public headers usable after only linking against the library target that the public header is sitting in.

All this was necessitated because I noticed in [traccc](https://github.com/acts-project/traccc) that some [detray](https://github.com/acts-project/detray) targets are not set up correctly. :frowning: But introducing rigorous tests revealed some problems already in [vecmem](https://github.com/acts-project/vecmem), and now in this project as well. (The `algebra::eigen_math`, `algebra::fastor_math`, `algebra::fastor_storage` and `algebra::vc_math` targets have some problems currently.)

Note that here I just made sure that the CMake configuration would correctly reflect what the code is doing. But for instance I'm not convinced that `algebra::vc_math` is indeed meant to rely on `algebra::cmath_math`. :confused:

https://github.com/acts-project/algebra-plugins/blob/main/math/vc/include/algebra/math/impl/vc_transform3.hpp#L11

Could you check Beomki and Joana?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ set( CMAKE_CUDA_EXTENSIONS FALSE CACHE BOOL "Disable (CUDA) C++ extensions" )
 set( CMAKE_SYCL_STANDARD 17 CACHE STRING "The (SYCL) C++ standard to use" )
 
 # Standard CMake include(s).
+include( CTest )
 include( GNUInstallDirs )
 
 # Flags controlling the meta-build system.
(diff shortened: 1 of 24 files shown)


krasznaa:
> The Vc `transform3` implementation relies on `cmath::cross`, it seems, not sure why it does not use `vc::math::cross`

It's because of:

```
[ 86%] Building CXX object tests/CMakeFiles/algebra_test_vc.dir/vc/vc_vc.cpp.o
In file included from /home/krasznaa/ATLAS/projects/algebra-plugins/algebra-plugins/tests/vc/vc_vc.cpp:12:
/home/krasznaa/ATLAS/projects/algebra-plugins/algebra-plugins/tests/common/test_host_basics.hpp:132:36: error: no matching function for call to 'cross'
  typename TypeParam::vector3 vG = algebra::vector::cross(vD, vF);
                                   ^~~~~~~~~~~~~~~~~~~~~~
...
/home/krasznaa/ATLAS/projects/algebra-plugins/algebra-plugins/math/vc/include/algebra/math/impl/vc_vector.hpp:66:33: note: candidate template ignored: substitution failure [with vector_type1 = typename TypeParam::vector3, vector_type2 = typename TypeParam::matrix<3, 1>]: invalid operands to binary expression ('value_type' (aka 'float') and 'value_type' (aka 'Vc_1::array<float, 3>'))
ALGEBRA_HOST_DEVICE inline auto cross(const vector_type1 &a,
                                ^
...
```

But I'd like to leave that can of worms to a separate PR...

PR #101, opened 2023-03-29
CMake Updates, main branch (2023.03.29.)

krasznaa:
Introduced `ALGEBRA_PLUGINS_FAIL_ON_WARNINGS`. Instead of only using `-Werror` (and similar flags) for debug builds, it is now possible to add `-Werror` to the build in any build mode. This is following the same sort of setup as in https://github.com/acts-project/vecmem/pull/210 and https://github.com/acts-project/traccc/pull/334.

Updated the CI to set `ALGEBRA_PLUGINS_FAIL_ON_WARNINGS` to `TRUE` in all tests.

Note that I was intending to add `-Wconversion` as a warning as well, but that triggered on a gazillion of floating point literals in the unit tests. (Complaining that double precision literals are converted into single precision values.) That just not seemed reasonable to do. As the readability of the unit tests would've been seriously hurt.

Also note that I realized, while updating the `builds.yml` configuration, that I forgot to add `-DALGEBRA_PLUGINS_BUILD_BENCHMARKS=TRUE` for the "native builds" in #99. So I'm sneaking in that fix here...

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -53,6 +53,8 @@ jobs:
                  -DALGEBRA_PLUGINS_INCLUDE_FASTOR=TRUE
                  -DALGEBRA_PLUGINS_SETUP_FASTOR=TRUE
                  -DALGEBRA_PLUGINS_USE_SYSTEM_FASTOR=FALSE
+                 -DALGEBRA_PLUGINS_BUILD_BENCHMARKS=TRUE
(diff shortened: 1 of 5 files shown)


niermann999 (approved):

niermann999:
I just used ```float``` literals in detray. Except for one test in the material, the testing accuracy seemed fine

PR #102, opened 2023-04-27
Added some more info to the README.

wermos:
I updated the docs to include information about the Fastor plugin and also made a couple of other miscellaneous tweaks.


wermos:
Also, I was unable to find the `ALGEBRA_PLUGINS_CUSTOM_SCALARTYPE` which is mentioned in the docs. Does it still exist, or should I remove that from the docs also in this PR?

wermos:
It looks like the failures are due to some issues with ghcr.io.

PR #103, opened 2023-05-02
Removed `ALGEBRA_PLUGINS_CUSTOM_SCALARTYPE` option from the README.

wermos:
I removed it because it looks like the current `CMakeLists.txt` does not use this option. It appears to have been removed in f4701ab13bd578eb5a486b65d1c461497573ece0 as a part of #44.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,6 @@ cmake --build algebra-plugins-build
 
 Available options:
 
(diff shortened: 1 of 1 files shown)


beomki-yeo (approved):
Looks good to me.
We are not using this option for CMake configuration because the scalar type is defined by algorithm itself.

PR #104, opened 2023-05-25
External Build Re-Design, main branch (2023.05.25.)

krasznaa:
I'm absolutely opening a can of worms with this one, but let's see where it goes...

With [CMake 3.24](https://cmake.org/cmake/help/latest/release/3.24.html) a new behaviour was introduced for [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) and [find_package](https://cmake.org/cmake/help/latest/command/find_package.html). They now implement a very similar logic for building or finding a needed external to what we implemented in these R&D projects. (And in [Acts](https://github.com/acts-project/acts) itself.)

Basically, [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) can either build an external dependency as it did before, or it can look for it with [find_package](https://cmake.org/cmake/help/latest/command/find_package.html). The default behaviour of the code is to silently look for the package first, and if it's found in the user's build environment, the found version is used. While if it's not found, it is built instead. But one can also force the build to either always look for an existing external, or to always build it. See: [FETCHCONTENT_TRY_FIND_PACKAGE_MODE](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE)

The behaviour of the code is not really changed with all these modifications. The reasons that doing these modifications could be a good idea are the following:
  - I like that in this setup we need to put [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) calls explicitly in all places in the code where we want to use an external dependency. Our current setup of only declaring those dependencies on the project level is not all too great...
  - Instead of using our own custom cache variables for controlling how externals would be used, we switch to using "CMake standard" variables for the same thing.
  - We don't have to worry about only a single project using [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) on let's say [Eigen](https://eigen.tuxfamily.org/) when building [traccc](https://github.com/acts-project/traccc). The resolution between "whose version wins out" is sort of well defined, so we don't need the logic implemented through variables like `ALGEBRA_PLUGINS_SETUP_EIGEN3` anymore.

But one of the main things that sold this for me is how one would be able to work on multiple projects at the same time in the new setup. With our current setup one needs to for instance:
  - Set up [vecmem](https://github.com/acts-project/vecmem) in its own directory. Set up its build in a separate directory, and install it into some directory.
  - With this project point to that installation location, and build against that location.
  - When modifying something in [vecmem](https://github.com/acts-project/vecmem), go through a full "build vecmem" -> "install vecmem" -> "reconfigure algebra-plugins" -> "build algebra-plugins" cycle.

In this new setup one can instead clone [vecmem](https://github.com/acts-project/vecmem) and this project side-by-side, and when setting up the build of this project, use `-DFETCHCONTENT_SOURCE_DIR_VECMEM=<foo>` to tell the build to pick up [vecmem](https://github.com/acts-project/vecmem) from that location. So all of a sudden one can modify both projects freely, and just issue `make` commands in a single build directory.

**But...** Because there is a big but... :frowning: We don't use CMake 3.24+ in our Docker images at the moment. :frowning: So to make all this work, we would need to put in a fair amount of effort... :thinking: So we may want to think it through properly before committing to it. Even if I think this could be helpful in the long term.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,9 +19,6 @@ set( CMAKE_SYCL_STANDARD 17 CACHE STRING "The (SYCL) C++ standard to use" )
 include( CTest )
 include( GNUInstallDirs )
 
-# Flags controlling the meta-build system.
-option( ALGEBRA_PLUGINS_USE_SYSTEM_LIBS "Use system libraries be default" FALSE )
-
 # Explicitly set the output directory for the binaries. Such that if this
 # project is included by another project, the main project's configuration would
 # win out.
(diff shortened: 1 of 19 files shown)


stephenswat (commented):
I like the fact that this reduces the number of configuration parameters, but does this mean that it is no longer possible to disable the remote fetching of packages (because `ALGEBRA_PLUGINS_USE_SYSTEM_LIBS` is gone)? :worried:

krasznaa:
> I like the fact that this reduces the number of configuration parameters, but does this mean that it is no longer possible to disable the remote fetching of packages (because `ALGEBRA_PLUGINS_USE_SYSTEM_LIBS` is gone)? worried

The direct replacement for `-DALGEBRA_PLUGINS_USE_SYSTEM_LIBS=TRUE` is: `-DFETCHCONTENT_TRY_FIND_PACKAGE_MODE=ALWAYS`.

Though I myself am also worried a bit about the lack of individual settings. :frowning: We can only force this behaviour globally now. But with [FETCHCONTENT_SOURCE_DIR_<FOO>](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_SOURCE_DIR_%3CuppercaseName%3E) I believe we would still have enough flexibility for all the use cases that we had so far.

wermos:
> The direct replacement for `-DALGEBRA_PLUGINS_USE_SYSTEM_LIBS=TRUE` is: `-DFETCHCONTENT_TRY_FIND_PACKAGE_MODE=ALWAYS.`

You can still keep the option `ALGEBRA_PLUGINS_USE_SYSTEM_LIBS` by simply setting the `FETCHCONTENT_TRY_FIND_PACKAGE_MODE` to `ALWAYS` whenever the former is passed on the command line 🤔

PR #106, opened 2023-08-29
Rollback to eigen 3.4.0

beomki-yeo:
Let me revert the eigen version as it turned out that 3.4.1 tag actually doesn't exist


beomki-yeo:
Let me merge this one...

PR #108, opened 2023-10-06
SYCL FP64 Test Skips, main branch (2023.10.06.)

krasznaa:
Some of the Intel (i)GPUs don't have native FP64 support. This update makes sure that tests would not fail on them because of this.

Without these checks, I would get the following on my laptop:

```
[ RUN      ] algebra_plugins/test_sycl_basics/sycl_array_cmath<double>.transform3
unknown file: error: C++ exception with description "Required aspect fp64 is not supported on the device" thrown in the test body.        

[  FAILED  ] algebra_plugins/test_sycl_basics/sycl_array_cmath<double>.transform3, ...
```

### DIFF
--- a/tests/accelerator/sycl/common/test_sycl_basics.hpp
+++ b/tests/accelerator/sycl/common/test_sycl_basics.hpp
@@ -1,6 +1,6 @@
 /** Algebra plugins library, part of the ACTS project
  *
- * (c) 2020-2022 CERN for the benefit of the ACTS project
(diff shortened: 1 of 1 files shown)


krasznaa:
Argh.. I was afraid that I would not find the right format on Windows... (Unfortunately running `clang-format` on Windows is quite impossible because of line-end-character differences. So Docker's `clang-format` does not have a good time there...)

beomki-yeo (approved):

PR #109, opened 2023-10-06
CMake Presets, main branch (2023.10.06.)

krasznaa:
Add a `CMakePresets.json` file to the repository. And also updated the CI to make use of it, simplifying the latter a little.

Presets are described here:

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

I've been thinking for a long time about adopting them in our projects to simplify builds a little bit. Since especially in this project (as @niermann999 has also rightfully complained in the past) we need to use an unreasonable number of CMake CACHE variables to specify exactly what type of build we want. With this setup the idea is that we could do:

```
cmake --preset default ../algebra-plugins/
```

Or:

```
cmake --preset cuda ../algebra-plugins/
```

Etc.

Note that it's very possible to use CACHE variable declarations together with the presets. Like:

```
cmake --preset default -DALGEBRA_PLUGINS_INCLUDE_SMATRIX=TRUE ../algebra-plugins/
```

I had to update the CI's Docker image versions, as the ones currently in use do not have a new enough version of CMake for this functionality. But as you can see, the new SYCL image triggers a numerical difference in some tests. :confused: So that will need to be sorted out before this could be merged in. :thinking:

P.S. I also added a previously missing "separate" Fastor test.

### DIFF
--- a/.github/ci_setup.sh
+++ b/.github/ci_setup.sh
@@ -1,6 +1,6 @@
 # Algebra plugins library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2022 CERN for the benefit of the ACTS project
+# (c) 2021-2023 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 #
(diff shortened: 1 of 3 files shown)


krasznaa:
Interestingly enough on my AMD CPU I could not reproduce the issue. :thinking: But I'm pretty sure that the problem came from the `ci_setup.sh` script needing some minor changes for the new SYCL Docker image.

niermann999 (approved):

PR #110, opened 2023-10-09
Silence Eigen Warnings From CUDA on Windows, main branch (2023.10.09.)

krasznaa:
Turned off specific CUDA warnings for the Eigen headers.

This is only needed with MSVC, since "system include directories" as a concept doesn't quite work with that compiler. 😦 (On Linux we are shielded from these warnings by: https://github.com/acts-project/algebra-plugins/blob/main/extern/eigen3/CMakeLists.txt#L39-L44) So we need to explicitly turn off the checking for some warnings in the code like this.

@niermann999, you mentioned in https://github.com/acts-project/algebra-plugins/pull/95#issuecomment-1752741760 that you observe warnings yourself as well. Could you elaborate? Since I only know at this point why there would be warnings on Windows. But on Linux there should not have been any to begin with. 🤔

P.S. Note that the `master` branch of Eigen doesn't produce any warnings/errors with our code. But it's not clear when we should expect new releases out of those guys. 😦 See:
  - https://gitlab.com/libeigen/eigen/-/issues/2699
  - https://gitlab.com/libeigen/eigen/-/issues/2724
  - etc.

### DIFF
--- a/math/eigen/include/algebra/math/impl/eigen_getter.hpp
+++ b/math/eigen/include/algebra/math/impl/eigen_getter.hpp
@@ -15,10 +15,17 @@
 #ifdef _MSC_VER
 #pragma warning(push, 0)
 #endif  // MSVC
+#ifdef __NVCC_DIAG_PRAGMA_SUPPORT__
+#pragma nv_diagnostic push
+#pragma nv_diag_suppress 20012
+#endif  // __NVCC_DIAG_PRAGMA_SUPPORT__
 #include <Eigen/Core>
 #ifdef _MSC_VER
 #pragma warning(pop)
 #endif  // MSVC
+#ifdef __NVCC_DIAG_PRAGMA_SUPPORT__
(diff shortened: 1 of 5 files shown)


niermann999:
I saw warnings in host code about uninitialized parameters that don't appear in the detray CI, but let me check again

beomki-yeo (approved):
Looks good to me. 

BTW, how do you compile and test MSVC compiler? Do you have any Windows image in your local linux machine or is that even possible

niermann999 (approved):

krasznaa:
> Looks good to me.
> 
> BTW, how do you compile and test MSVC compiler? Do you have any Windows image in your local linux machine or is that even possible

Windows image... I use Windows on my laptop for this. :wink:

image

beomki-yeo:
OK. CI failures with MSVC was a horror to me when I didn't have a Windows license :smiley:

PR #112, opened 2024-01-22
Make __host__ and __device__ function checks work with HIP

StewMH:
As proposed by @krasznaa in https://github.com/acts-project/vecmem/pull/261, extend the __host__ and __device__ macros for HIP.

### DIFF
--- a/common/include/algebra/qualifiers.hpp
+++ b/common/include/algebra/qualifiers.hpp
@@ -8,25 +8,25 @@
 
 #pragma once
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define ALGEBRA_DEVICE __device__
 #else
 #define ALGEBRA_DEVICE
 #endif
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define ALGEBRA_HOST __host__
 #else
 #define ALGEBRA_HOST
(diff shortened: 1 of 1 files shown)


common/include/algebra/qualifiers.hpp
```diff
@@ -8,25 +8,25 @@
...
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define ALGEBRA_HOST_DEVICE __host__ __device__
 #else
 #define ALGEBRA_HOST_DEVICE
 #endif
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define ALGEBRA_ALIGN(x) __align__(x)
```
krasznaa, line 30: Hopefully HIP also needs `__align__`. But I could just not figure this out with relatively simple Google searches. 😦 The other keywords (`__device__`, etc.) of course are universal between CUDA and HIP.

My educated guess is that they copied the CUDA formalism for `__align__` as well. 🤔 But we should remember this, in case we run into some issues later on.

krasznaa (approved):
Yeah, fully on board.

common/include/algebra/qualifiers.hpp
StewMH: Yeah I'm not sure either, but I guess we'll find out.

PR #120, opened 2024-04-20
Vectorize the rotate function of cmath transform3

beomki-yeo:
Vectorize the rotate function of cmath transform3
Also vectorized the `==` operator

### DIFF
--- a/math/cmath/include/algebra/math/impl/cmath_transform3.hpp
+++ b/math/cmath/include/algebra/math/impl/cmath_transform3.hpp
@@ -191,8 +191,8 @@ struct transform3 {
   ALGEBRA_HOST_DEVICE
   inline bool operator==(const transform3 &rhs) const {
 
-    for (size_type i = 0; i < 4; i++) {
-      for (size_type j = 0; j < 4; j++) {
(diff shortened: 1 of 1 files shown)


niermann999 (approved):

beomki-yeo:
@niermann999 will look into a detailed report from compiler tomorrow but is using the for loop mandatory for vectorization?

niermann999:
> @niermann999 will look into a detailed report from compiler tomorrow but is using the for loop mandatory for vectorization?

I don't know too much about autovectorization tbh (and if it is the same between different compilers). It seems though that there is SLP vectorization next to the loop vectorization, which should be capable of vectorizing simple code blocks where operations are repeated for multiple scalars.

beomki-yeo:
OK. 
It seems the SLP vectorization is enabled by default.
https://gcc.gnu.org/projects/tree-ssa/vectorization.html

> Basic block vectorization, aka SLP, is enabled by the flag -ftree-slp-vectorize, and requires the same platform dependent flags as loop vectorization. Basic block SLP is enabled by default at -O3 and when -ftree-vectorize is enabled.

PR #125, opened 2024-07-23
Disable Fastor in macOS CI job

stephenswat:
XCode refuses to compile algebra-plugins with the Fastor plugin turned on, and this is causing the CI to fail. This commit disables Fastor in the macOS CI job to ensure that it runs through.


stephenswat:
I think this closes #123.

niermann999 (approved):

PR #127, opened 2024-11-11
feat: matrix implementation for vc

niermann999:
Implement a matrix type for the AoS and SoA Vc types and add matrix benchmarks. The matrix type works the same way as the vector type that we already have: For AoS it uses the ```Vc::SimdArray``` as columns, for SoA it uses an ```std::array``` of ```Vc::Vectors``` in each element (hidden behind the ```storage::vector``` class). The basic matrix functionality is implemented such, that single elements of the columns are accessed as rarely as possible and we try to handle the entire column if possible. The vector class has been extended a little to make sure trailing entries that are there for padding are initialized to zero (for some reason that is performance relevant in vectorization)


niermann999:
Based on #128

math/vc_aos/include/algebra/math/impl/vc_aos_matrix.hpp
```diff
@@ -0,0 +1,127 @@
...
+    return determinant_t{}(m);
+  }
+
+  /// Get inverse using a specific algorithm @tparam inverse_t
+  template <std::size_t N, typename inverse_t = inverse_t>
+  ALGEBRA_HOST_DEVICE static constexpr constexpr matrix_type<N, N> inverse(
+      const matrix_type<N, N> &m) {
+
+    return inverse_t{}(m);
+  }*/
```
krasznaa, line 122: Should we add disabled lines in this PR?

math/vc_aos/include/algebra/math/impl/vc_aos_transform3.hpp
```diff
@@ -36,6 +36,14 @@ using algebra::storage::operator+;
 template <template <typename, std::size_t> class array_t, typename scalar_t>
 struct transform3 {
 
+  // How to address matrix columns
+  enum elem : std::size_t {
+    e_x = 0u,
+    e_y = 1u,
+    e_z = 2u,
+    e_t = 3u,
+  };
```
krasznaa, line 45: Should we make it private? I'm not too keen on introducing a public "non-class" enumeration. 🤔

math/vc_aos/include/algebra/math/impl/vc_aos_transform3.hpp
```diff
@@ -224,27 +239,27 @@ struct transform3 {
     array_type<16> submatrix;
     for (unsigned int irow = 0; irow < 3; ++irow) {
       for (unsigned int icol = 0; icol < 3; ++icol) {
-        submatrix[icol + irow * 4] = element_getter()(_data, irow, icol);
+        submatrix[icol + irow * 4] = _data[icol][irow];
```
krasznaa, line 242: Can't we use an appropriate element getter here still?

storage/common/include/algebra/storage/matrix.hpp
```diff
@@ -0,0 +1,454 @@
...
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+// @TODO: Remove this when Vc fixes their false positives.
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic warning "-Wdeprecated-declarations"
+#endif
```
krasznaa, line 12: I'm very baffled by this. This also seems to enable a warning, not disable it. Am I reading this wrong?

storage/common/include/algebra/storage/matrix.hpp
```diff
@@ -0,0 +1,454 @@
...
+          std::size_t ROW, std::size_t COL>
+struct alignas(alignof(storage::vector<ROW, value_t, array_t>)) matrix {
+
+  // The matrix consists of column vectors
+  using vector_type = storage::vector<ROW, value_t, array_t>;
+  // Value type: Can be simd types
+  using value_type = value_t;
+
+  /// Default constructor
+  constexpr matrix() = default;
```
krasznaa, line 34: Yay on the `constexpr`. But I would still use `ALGEBRA_HOST_AND_DEVICE`. So that clients would not be forced into using a (still) experimental feature with CUDA.

storage/common/include/algebra/storage/vector.hpp
```diff
@@ -20,45 +25,76 @@ namespace algebra {
...
+
   // Value type is a simd vector in SoA and a scalar in AoS
   using value_type = value_t;
   /// Underlying data array type
-  using array_type = array_t<value_t, N>;
+  using array_type = array_t<value_t, simd_size()>;
 
   /// Default contructor sets all entries to zero.
-  constexpr vector() { zero_fill(std::make_index_sequence<N>{}); }
+  constexpr vector() {
```
krasznaa, line 62: I would add a bunch of `ALGEBRA_HOST_AND_DEVICE` declarations here as well. 🤔

krasznaa (commented):
I'm generally in favour, as much as I can tell. (It's a lot of code...) Just some technical comments...

storage/common/include/algebra/storage/matrix.hpp
niermann999: If I use ```ALGEBRA_HOST_AND_DEVICE``` with a defaulted constructor CUDA will complain

math/vc_aos/include/algebra/math/impl/vc_aos_matrix.hpp
niermann999: I can remove them for now. They will be added in one of the other PRs

storage/common/include/algebra/storage/matrix.hpp
krasznaa: I should've opened the comment on a different function. (You're right, on this particular function you mustn't add the macro.)

But on all the other functions, I feel pretty strongly about `ALGEBRA_...`. Since this is in a common piece of code here. We should make it technically allowed to use this code on a GPU.

math/vc_aos/include/algebra/math/impl/vc_aos_transform3.hpp
niermann999: sure, done

math/vc_aos/include/algebra/math/impl/vc_aos_transform3.hpp
niermann999: I don't really see what that gains us here, when we can access the matrix natively the same way the getter does anyway? 

From what I can see, the getters are needed in generic code that can be shared by multiple plugins. In all other plugins except the cmath plugin we also access the data natively here instead of through an extra layer. And I find that easier to undestand tbh

storage/common/include/algebra/storage/matrix.hpp
niermann999: Some of this code uses vectorization specific functionality like boolean mask evaluation. Maybe we can hide that behind an interface such that it has a chance of ever compiling in device code?

stephenswat (approved):
Seems like all the issues have been addressed and there are a lot of PRs waiting for this, so let's go ahead with this and then we can iterate later if we want.

PR #128, opened 2024-11-12
ref: Update CI

niermann999:
Update github actions and switch to newer containers


stephenswat (approved):
LGTM!

.github/ci_setup.sh
```diff
@@ -21,5 +21,6 @@ if [ "${PLATFORM_NAME}" = "SYCL" ]; then
    export CC=`which clang`
    export CXX=`which clang++`
    export SYCLCXX="${CXX} -fsycl"
-   export SYCL_DEVICE_FILTER=host
+   export ONEAPI_DEVICE_SELECTOR=host
```
krasznaa, line 24: I don't think this will work... 🤔 You'll rather need:

```suggestion
   export ONEAPI_DEVICE_SELECTOR=opencl:cpu
```

Similar to for instance: https://github.com/acts-project/traccc/blob/main/.gitlab-ci.yml#L86

See: https://intel.github.io/llvm/EnvironmentVariables.html#oneapi-device-selector

niermann999: Changed

krasznaa:
Unfortunately Vc is beyond repair... 😦 Just turn it off in the Windows CI build. (`-DALGEBRA_PLUGINS_SETUP_VC=FALSE -DALGEBRA_PLUGINS_INCLUDE_VC=FALSE` with some YAML logic...)

I also see an error coming from Eigen on Windows, while building the CUDA tests. Which don't actually look like an MSVC bug, but rather an Eigen one. But since the CI doesn't test CUDA on Windows, let's just ignore that one for now... 🤔

niermann999:
MSVC seems fine now, but I see lot's of segfaults in the SYCL tests....

niermann999:
I think the sycl problem here is the same as https://github.com/acts-project/traccc/issues/655, so I have disabled the failing tests for now

.github/ci_setup.sh
```diff
@@ -15,11 +15,14 @@ PLATFORM_NAME=$1
 
 # Set up the correct environment for the SYCL tests.
 if [ "${PLATFORM_NAME}" = "SYCL" ]; then
-   source /opt/intel/oneapi/setvars.sh --include-intel-llvm
+   if [ -f "/opt/intel/oneapi/setvars.sh" ]; then
+      source /opt/intel/oneapi/setvars.sh --include-intel-llvm
+   fi
```
krasznaa, line 20: Curious. This is not a bad addition per se, but does one of our current oneAPI images not have oneAPI installed under `/opt/intel/oneapi`? 😕

But sure, we may very well want to add SYCL builds not based on oneAPI at one point. 🤔

.github/workflows/builds.yml
```diff
@@ -62,28 +69,28 @@ jobs:
...
-            CONTAINER: "ghcr.io/acts-project/ubuntu2004:v43"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404:56"
             OPTIONS: --preset fastor
           - NAME: "CUDA"
-            CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:v43"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404_cuda:67"
             OPTIONS: --preset cuda
           - NAME: "SYCL"
-            CONTAINER: "ghcr.io/acts-project/ubuntu2004_oneapi:v43"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404_oneapi:56"
```
krasznaa, line 93: Why not use the same (`67`) tag for all of these?

And while at it, could you add SYCL builds with `"ghcr.io/acts-project/ubuntu2404_cuda_oneapi:67"` and `"ghcr.io/acts-project/ubuntu2404_rocm_oneapi:67"` as well? I'd be curious to see if those succeed... 🤔

tests/accelerator/sycl/common/test_sycl_basics.hpp
```diff
@@ -165,5 +165,6 @@ TYPED_TEST_P(test_sycl_basics, transform3) {
   this->compareOutputs();
 }
 
-REGISTER_TYPED_TEST_SUITE_P(test_sycl_basics, vector_2d_ops, vector_3d_ops,
-                            matrix64_ops, matrix22_ops, transform3);
+REGISTER_TYPED_TEST_SUITE_P(test_sycl_basics, /*vector_2d_ops, vector_3d_ops,
+                            matrix64_ops, matrix22_ops,*/
+                            transform3);
```
krasznaa, line 170: How about, instead of turning off the build, turning off the running of the tests instead? 🤔 I would still like to keep the build tests alive...

krasznaa (commented):
I think I'll use this repo to get help from the Intel developers for our issues. So I'm very much on board with the C\+\+20 upgrade. I just don't want to turn off more than absolutely needed. 🤔

.github/workflows/builds.yml
niermann999: Updated all to 68

.github/ci_setup.sh
```diff
@@ -15,11 +15,14 @@ PLATFORM_NAME=$1
...
    # Use clang/clang++ instead of icx/icpx, to avoid some aggressive math
    # optimizations that break some tests.
    export CC=`which clang`
    export CXX=`which clang++`
    export SYCLCXX="${CXX} -fsycl"
-   export SYCL_DEVICE_FILTER=host
+   export ONEAPI_DEVICE_SELECTOR=opencl:cpu
+   export SYCLFLAGS="-fsycl -fsycl-targets=spir64"
```
krasznaa, line 27: Of course the problem is that now that you added NVIDIA and AMD backended build jobs, this setting mustn't be done here. Since right now we continue to only do Intel backed builds in the CI, with 3 different Docker images. 😛

niermann999:
Also updated the vecmem version now, since it was before only done on one of the dependent PRs

.github/ci_setup.sh
stephenswat, line 27: The unfortunate thing is that this will now override the CUDA and ROCm SYCL targets in the new Docker images, so you'll need to find a different way to do this.

.github/workflows/builds.yml
```diff
@@ -62,29 +69,45 @@ jobs:
...
           - NAME: "HOST"
-            CONTAINER: "ghcr.io/acts-project/ubuntu2004:v43"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404:68"
             OPTIONS: --preset eigen -DALGEBRA_PLUGINS_USE_SYSTEM_EIGEN3=TRUE
+            run_tests: true
           - NAME: "HOST"
-            CONTAINER: "ghcr.io/acts-project/ubuntu2004:v43"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404:68"
             OPTIONS: --preset eigen
+            run_tests: true
```
krasznaa, line 78: Let's follow the capitalization used in all other variables. I.e. let's call this `RUN_TESTS`.

.github/ci_setup.sh
niermann999: Can I pass this as DSYCLFLAGS in the yaml?

.github/workflows/builds.yml
```diff
@@ -62,29 +69,45 @@ jobs:
...
+            OPTIONS: --preset sycl
+            run_tests: false
+          - NAME: "SYCL CUDA"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404_cuda_oneapi:68"
+            OPTIONS: --preset sycl
+            run_tests: false
+          - NAME: "SYCL ROCM"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404_rocm_oneapi:68"
             OPTIONS: --preset sycl
+            run_tests: false
```
krasznaa, line 110: As @stephenswat pointed out as well, we should add something like an `ENV` variable, which the CI steps could use to set environment variables that are specific to the different jobs. This would allow moving the lines from `ci_setup.sh` here.

.github/ci_setup.sh
stephenswat: I think it's `-DCMAKE_SYCL_FLAGS`. But yes.

krasznaa: That would technically work. But `ONEAPI_DEVICE_SELECTOR` is strictly an environment variable. We should not set a uniform value for all backends for that...

stephenswat: But `ONEAPI_DEVICE_SELECTOR` is only for runtime, so that doesn't actually need to be changed as we are not actually running the CUDA + ROCm code (yet!).

niermann999: OK, can we do that in a separate PR maybe? Because I have a lot of work depending on this PR...

.github/workflows/builds.yml
niermann999: Done

stephenswat (approved):

tests/accelerator/sycl/common/test_sycl_basics.hpp
niermann999: Done

krasznaa (approved):
Fine, let's finally go ahead with it. I'd also like to move on with further debugging the SYCL issues...

niermann999:
I can't merge this anymore. Can somebody hit the button for me @krasznaa @paulgessinger ?

krasznaa:
> I can't merge this anymore. Can somebody hit the button for me @krasznaa @paulgessinger ?

Unfortunately in yesterday's cleanup I also lost my privileges on the repository. 😦

image

.github/workflows/builds.yml
niermann999: Went back to 56 as some additional ROOT config would be needed with the latest containers

paulgessinger:
Whoops, let me check what's going on.

paulgessinger:
@krasznaa @niermann999 can you try again now?

krasznaa:
Yepp. Looks better now. 😉

image

PR #131, opened 2024-11-16
feat: generic plugin

niermann999:
The cmath frontend cannot be used generically with every linear algebra library, since the subscript operator is not always available. In the current matrix algorithm implementation, a more generic solution is being used by applying the element getter instead. In this PR, this was generalized into a *generic* implementation, which can be used on top of every linear algebra plugin. The cmath, which relies on the c-style access to arrays, implementation is kept as well, since it is still shared between the ```array``` and ```vecmem``` plugins.


niermann999:
Based on #130

stephenswat:
Is this ready for review? And if so - any guidelines on how to do this? In other words, are there any major changes here or is it mostly renaming stuff?

niermann999:
> Is this ready for review? And if so - any guidelines on how to do this? In other words, are there any major changes here or is it mostly renaming stuff?

Yes, I think a lot of it is renaming and moving code around. The getters, for example, were moved into their own files to be able to combine them with different matrix implementations

stephenswat (approved):
I'm not going to look through all of this, but it looks sensible; happy to scream test this.

PR #132, opened 2024-11-16
feat: Introduce concepts

niermann999:
This PR adds concepts for the main linear algebra types and constrains as many template parameters as possible. Also switches out the ```std::enable_if``` instances against ```requires``` clauses

### DIFF
--- a/benchmarks/vc_soa/include/benchmark/vc_soa/data_generator.hpp
+++ b/benchmarks/vc_soa/include/benchmark/vc_soa/data_generator.hpp
@@ -17,11 +17,11 @@
 namespace algebra {
 
 /// Fill a @c Vc::Vector based vector with random values
-template <typename vector_soa_t>
+template <concepts::vector vector_soa_t>
 inline void fill_random_vec(std::vector<vector_soa_t> &collection) {
   // Generate a vector of the right type with random values
   auto rand_obj = []() {
-    using simd_vector_t = typename vector_soa_t::value_type;
+    using simd_vector_t = typename vector_soa_t::scalar_type;
     vector_soa_t tmp{};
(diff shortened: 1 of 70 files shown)


niermann999:
Based on #131

benchmarks/vc_soa/include/benchmark/vc_soa/data_generator.hpp
```diff
@@ -80,8 +80,9 @@ inline void fill_random_matrix(std::vector<matrix_t> &collection) {
 
       for (std::size_t i = 0u; i < matrix_t::rows(); ++i) {
         v[i] = simd_vector_t::Random();
-        m[j] = v;
       }
+
+      m[j] = v;
```
krasznaa, line 85: Oh, snap... Did you find a bug as part of these updates? 🤔

krasznaa (approved):
Real talk: I did not read through every single line. 😦 But the things that I did read through, I liked. So I'm here for this PR. 👍

benchmarks/vc_soa/include/benchmark/vc_soa/data_generator.hpp
niermann999: This is part of the random data generator for the benchmarks. And I think it has been setting the same vector multiple times instead of once, so it was just doing some pointless work

krasznaa: 👍 I saw that this was in the benchmarks. That's why I was also not super worried about it. 😉

krasznaa:
Did you rebase on top of the latest SYCL changes? I fear that this repository still doesn't require PRs to have run their CI on top of the tip of the `main` branch. 🤔

PR #133, opened 2024-11-18
ref: Adapt to client interface and allow to compile multiple plugins side by side

niermann999:
Hide the generic function overloads behind the functions of specific plugins, because they would otherwise always take precedence when using multiple plugins side by side.

### DIFF
--- a/common/include/algebra/type_traits.hpp
+++ b/common/include/algebra/type_traits.hpp
@@ -81,8 +81,7 @@ struct dimensions {
 
 /// Specilization for scalar types
 template <typename M>
-  requires std::is_fundamental_v<M>
-struct dimensions<M> {
+requires std::is_fundamental_v<M> struct dimensions<M> {
 
   using size_type = std::size_t;
 
@@ -138,11 +137,11 @@ template <typename T>
 struct get_algebra {};
 
 template <typename T>
-  requires(!std::is_same_v<typename T::point3D, void>)
-struct get_algebra<T> {
+requires(!std::is_same_v<typename T::point3D, void>) struct get_algebra<T> {
   template <typename U>
   using simd = typename T::template simd<U>;
   using size_type = typename T::size_type;
+  using boolean = typename T::boolean;
   using value = typename T::value_type;
   using scalar = typename T::scalar;
   using point2D = typename T::point2D;
@@ -156,10 +155,19 @@ struct get_algebra<T> {
 }  // namespace traits
 
 template <typename A>
-using get_scalar_t = typename traits::get_algebra<A>::scalar;
+using get_value_t = typename traits::get_algebra<A>::value;
+
+template <typename A>
+using get_boolean_t = typename traits::get_algebra<A>::boolean;
 
 template <typename A, typename T>
 using get_simd_t = typename traits::get_algebra<A>::template simd<T>;
+
+template <typename A>
+using get_size_t = typename traits::get_algebra<A>::size_type;
+
+template <typename A>
+using get_scalar_t = typename traits::get_algebra<A>::scalar;
 
 template <typename A>
 using get_point2D_t = typename traits::get_algebra<A>::point2D;
@@ -172,9 +180,6 @@ using get_vector3D_t = typename traits::get_algebra<A>::vector3D;
 
 template <typename A>
 using get_transform3D_t = typename traits::get_algebra<A>::transform3D;
-
-template <typename A>
-using get_size_t = typename traits::get_algebra<A>::size_type;
 
 template <typename A, std::size_t R, std::size_t C>
 using get_matrix_t = typename traits::get_algebra<A>::template matrix<R, C>;
--- a/common/include/algebra/type_traits.hpp
+++ b/common/include/algebra/type_traits.hpp
@@ -12,7 +12,9 @@
 #include <cmath>
 #include <type_traits>
 
-namespace algebra::traits {
+namespace algebra {
+
+namespace traits {
 
 /// Matrix traits
 /// @{
@@ -79,7 +81,8 @@ struct dimensions {
 
 /// Specilization for scalar types
 template <typename M>
-requires std::is_fundamental_v<M> struct dimensions<M> {
+  requires std::is_fundamental_v<M>
+struct dimensions<M> {
 
   using size_type = std::size_t;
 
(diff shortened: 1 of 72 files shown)


niermann999:
Based on #132

krasznaa:
Let's merge #132 first! That should make it a lot easier (for me) to review this one... 🤔

common/include/algebra/type_traits.hpp
```diff
@@ -129,7 +131,61 @@ template <class M>
...
+  using vector3D = typename T::vector3D;
+  using transform3D = typename T::transform3D;
+  template <std::size_t ROWS, std::size_t COLS>
+  using matrix = typename T::template matrix<ROWS, COLS>;
+};
+
+}  // namespace traits
+
+template <typename A>
+using get_value_t = typename traits::get_algebra<A>::value;
```
krasznaa, line 158: Should we use the `get_` prefix for all these things? I'm not sure myself...

frontend/array_cmath/include/algebra/array_cmath.hpp
```diff
@@ -10,6 +10,7 @@
 // Project include(s).
 #include "algebra/math/cmath.hpp"
 #include "algebra/math/generic.hpp"
+#include "algebra/print.hpp"
```
krasznaa, line 13: I don't love this. 😦

If you insist on adding print helpers, let's so that in a separate library. Something like `algebra::utils`. I really don't want this code to be always there, even when we're not interested in printing these types.

frontend/array_cmath/include/algebra/array_cmath.hpp
```diff
@@ -110,4 +115,31 @@ using transform3 =
 
 }  // namespace array
 
+namespace plugin {
+
+/// Define the plugin types
+/// @{
+template <concepts::value V>
+struct cmath {
```
krasznaa, line 123: Given the library and file names, should this not rather be called `array_cmath`?

Given that it's really "just" the storage that is getting defined, if not `array_cmath`, it would have to be `array`. 🤔

krasznaa (changes requested):
I have some hang-ups about this one. 🤔

common/include/algebra/type_traits.hpp
```diff
@@ -129,7 +131,61 @@ template <class M>
...
 /// @}
 
-}  // namespace algebra::traits
+/// The algebra types
+/// @{
+template <typename T>
+struct get_algebra {};
+
+template <typename T>
+requires(!std::is_same_v<typename T::point3D, void>) struct get_algebra<T> {
```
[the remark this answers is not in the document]
niermann999: The algebra concepts use the algebra type traits, I am getting into a circular dependency otherwise

common/include/algebra/type_traits.hpp
niermann999: I don't know either tbh. They are "typedefed away" in the detray namespace, so nobody will really see them I think

frontend/array_cmath/include/algebra/array_cmath.hpp
niermann999: renamed

frontend/array_cmath/include/algebra/array_cmath.hpp
niermann999: Moved to ```algebra_utils```. The plan was to use them in detray core though, otherwise it is not really useful for logging and debugging

krasznaa: `detray::core`? 😕 But `std::ostream` is not something that would ever work in GPU code. If you need logging capabilities in Detray, this is **not** the interface for doing it. 🤔

niermann999: True, but the code is fully heterogeneous, so I for myself do all of the logging (using the inspectors), as well as adhoc debugging on CPU, where the stream operators are very handy. Almost every type in detray comes with such an operator already

math/generic/include/algebra/math/generic.hpp
```diff
@@ -8,6 +8,8 @@
 #pragma once
 
 // Impl include(s).
+#include "algebra/math/boolean.hpp"
+#include "algebra/math/common.hpp"
```
krasznaa, line 12: Same question here. Why do we need `boolean.hpp` here?

I assume `common.hpp` was missing earlier, right? Since none of the current changes justify its addition either.

math/vc_soa/include/algebra/math/impl/vc_soa_math.hpp
```diff
@@ -0,0 +1,151 @@
...
+#include <Vc/Vc>
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif  // MSVC
+
+// System include(s)
+#include <algorithm>
+#include <cmath>
+
+namespace algebra::math_soa {
```
krasznaa, line 23: Also here. Wouldn't it be easier to just use `algebra::math` as-is?

math/vc_soa/include/algebra/math/impl/vc_soa_math.hpp
```diff
@@ -0,0 +1,151 @@
...
+using std::sinh;
+using std::sqrt;
+using std::tan;
+using std::tanh;
+/// @}
+
+/// Vc overloads of common math functions
+/// @{
+template <typename T>
+requires Vc::Traits::is_simd_vector<T>::value inline decltype(auto) abs(
```
krasznaa, line 58: I'm asking this as a concept newbie: Would it not be nicer to define a concept with the Vc type trait, and use that in this declaration instead?

krasznaa (commented):
I still like where this is heading, just some more technical questions to slow things down...

math/cmath/include/algebra/math/impl/cmath_vector.hpp
```diff
@@ -36,18 +119,11 @@ requires std::is_scalar_v<scalar_t> ALGEBRA_HOST_DEVICE inline scalar_t dot(
   return ret;
 }
 
-/// Dot product between two input vectors
-///
-/// @param a the first input vector
-/// @param b the second input matrix with single column
-///
-/// @return the scalar dot product value
```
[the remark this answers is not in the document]
niermann999: I put the functions between a /// @{ /// @} because they all do the same thing essentially. Or is this not applying a single copy of the comment to all of them?

math/vc_soa/include/algebra/math/impl/vc_soa_boolean.hpp
```diff
@@ -0,0 +1,49 @@
...
+// Vc include(s).
+#ifdef _MSC_VER
+#pragma warning(push, 0)
+#endif  // MSVC
+#include <Vc/Vc>
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif  // MSVC
+
+namespace algebra::boolean_soa {
```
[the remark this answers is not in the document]
niermann999: Tbh, I am not sure. I had a hard time making detray find the overloads, let me see if it works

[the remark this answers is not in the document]
niermann999: Seems to work, done.

math/vc_soa/include/algebra/math/impl/vc_soa_math.hpp
niermann999: Done

PR #135, opened 2024-11-18
Update deprecated GitHub Action versions

stephenswat:
Several of our GitHub Actions are deprecated, causing the CI to automatically fail. This commit updates the relevant jobs.


niermann999:
I think this is already part of #128. Just waiting on the green lights regarding to switch off the SYCL tests temporarily @krasznaa ?

stephenswat:
Ah good, sorry; I hadn't seen that!

PR #137, opened 2024-11-19
Add new generic matrix operators

stephenswat:
Our current approach for e.g. matrix multiplication reads very naturally, e.g. `A = B*C` models $A = BC$, but this has a problem on GPU code. Indeed, if these matrices have size $N \times N$, then we first concretize an $N \times N$ matrix which we then have to copy element-by-element into $A$. This means that we need to keep that many registers live, which is quite large for e.g. $7 \times 7$ free matrices (which thus occupy 49 registers).

This problem can be alleviated by implementing optimized operators. More precisely, this PR implements the following new methods:

* `set_product(A, B, C)` computes $A = BC$ without intermediate values.
* `set_product_left_transpose(A, B, C)` computes $A = B^TC$ without intermediate values.
* `set_product_right_transpose(A, B, C)` computes $A = BC^T$ without intermediate values.
* `set_inplace_product_right(A, B)` computes $A = AB$ in place.
* `set_inplace_product_left(A, B)` computes $A = BA$ in place.
* `set_inplace_product_right_transpose(A, B)` computes $A = AB^T$ in place.
* `set_inplace_product_left_transpose(A, B)` computes $A = B^TA$ in place.

Includes tests; depends on #134.

### DIFF
--- a/common/include/algebra/concepts.hpp
+++ b/common/include/algebra/concepts.hpp
@@ -88,6 +88,34 @@ concept column_matrix = matrix<M> && (algebra::traits::columns<M> == 1);
 
 template <typename M>
 concept column_matrix3D = column_matrix<M> && (algebra::traits::rows<M> == 3);
+
+template <typename MA, typename MB>
+concept matrix_compatible = matrix<MA>&& matrix<MB>&& std::convertible_to<
+    algebra::traits::index_t<MA>, algebra::traits::index_t<MB>>&&
+    std::convertible_to<algebra::traits::index_t<MB>,
+                        algebra::traits::index_t<MA>>;
+
+template <typename MA, typename MB>
+concept matrix_multipliable =
+    matrix_compatible<MA, MB> &&
+    (algebra::traits::columns<MA> ==
+     algebra::traits::rows<MB>)&&requires(algebra::traits::scalar_t<MA> sa,
+                                          algebra::traits::scalar_t<MB> sb) {
+  {(sa * sb) + (sa * sb)};
+};
+
+template <typename MA, typename MB, typename MC>
+concept matrix_multipliable_into =
+    matrix_multipliable<MA, MB>&& matrix_compatible<MA, MC>&&
+        matrix_compatible<MB, MC> &&
+    (algebra::traits::rows<MC> == algebra::traits::rows<MA>)&&(
+        algebra::traits::columns<MC> ==
+        algebra::traits::columns<
+            MB>)&&requires(algebra::traits::scalar_t<MA> sa,
+                           algebra::traits::scalar_t<MB> sb,
+                           algebra::traits::scalar_t<MC>& sc) {
+  {sc += (sa * sb)};
+};
 /// @}
 
 /// Transform concept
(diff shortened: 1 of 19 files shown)


stephenswat:
I was considering making the translate on the left- and right-hand size a boolean parameter or even a boolean template, but I wasn't sure... :thinking:

stephenswat:
@niermann999 I think the generic plugin (#131) would actually be super useful here; do you think we can incorporate this after merging the generic plugin?

niermann999:
> @niermann999 I think the generic plugin (#131) would actually be super useful here; do you think we can incorporate this after merging the generic plugin?

Yes, sure. It should be possible to use this with any linear algebra backend, also the vectorized ones (I did this in the last PR with the determinant and inverse now, until I had time to look into optimizations). Unfortunately, the full implementation of the generic plugin kind of happened on the way, so it is three PRs down the pipeline...

stephenswat:
@niermann999 do you have some ideas on how I would now move this to your new generic algebra plugin?

niermann999:
Can you also try to add this operation to the matrix benchmarks?

beomki-yeo:
I am onboard with the direction but could you make this operators more CPU-vectorization friendly?

stephenswat:
> I am onboard with the direction but could you make this operators more CPU-vectorization friendly?

Hi Beomki, thanks for the comment but I am not sure which of the functions you mean; where do you see any particular vectorization unfriendliness?

stephenswat:
@niermann999 this is now ready for review.

common/include/algebra/concepts.hpp
```diff
@@ -88,6 +88,41 @@ concept column_matrix = matrix<M> && (algebra::traits::columns<M> == 1);
 
 template <typename M>
 concept column_matrix3D = column_matrix<M> && (algebra::traits::rows<M> == 3);
+
+template <typename MA, typename MB>
+concept matrix_multipliable =
+    matrix<MA>&& matrix<MB> &&
+    (algebra::traits::columns<MA> ==
+     algebra::traits::rows<
+         MB>)&&std::convertible_to<algebra::traits::index_t<MA>,
```
niermann999, line 97: Should we maybe factor a part of this concept into a smaller concept like e.g. ```compatible_matrix```, where we test the index and scalar types?

tests/common/test_host_basics.hpp
```diff
@@ -425,9 +718,124 @@ TYPED_TEST_P(test_host_basics_matrix, matrix_3x3) {
     ASSERT_NEAR(algebra::getter::element(m33_inv, 2, 2), -10.f / 20.f,
                 this->m_isclose);
   }
+
+  {
+    typename TypeParam::template matrix<N, N> m1;
```
niermann999, line 723: Is this the same test code as below? Can this be written as a generic function or is that not possible?

stephenswat: Great catch; I've deduplicated the testing code.

stephenswat:
Ah yes, more obscure MSVC clownery. :clown_face:

stephenswat:
@niermann999 finally passes the CI; what do you think?

niermann999 (approved):

PR #140, opened 2024-12-04
Feat: Dedicated row and column getters

niermann999:
Get vector types from matrices directly


stephenswat:
Closing in order to archive this repository.

PR #141, opened 2024-12-04
fix: Eigen Identity matrix and Vc/SMatrix matrix getter and hide arithmetic operators in algebra namespace

niermann999:
The Eigen Identity matrix in the transform class leads to invalid instructions in CUDA. Also fixes small issues in the matrix getters of the Vc and SMatrix plugins.

Also moves the arithmetic operators of the cmath plugin to the algebra namespace, since they were picked up in ACTS builds by Eigen types

### DIFF
--- a/math/eigen/include/algebra/math/impl/eigen_transform3.hpp
+++ b/math/eigen/include/algebra/math/impl/eigen_transform3.hpp
@@ -163,11 +163,12 @@ struct transform3 {
   /// @param m is the rotation matrix
   /// @param v is the vector to be rotated
   template <typename derived_type>
-    requires(Eigen::MatrixBase<derived_type>::RowsAtCompileTime == 3 &&
-             Eigen::MatrixBase<derived_type>::ColsAtCompileTime == 1)
-  ALGEBRA_HOST_DEVICE static inline auto rotate(
-      const Eigen::Transform<scalar_type, 3, Eigen::Affine> &m,
-      const Eigen::MatrixBase<derived_type> &v) {
+  requires(Eigen::MatrixBase<derived_type>::RowsAtCompileTime == 3 &&
+           Eigen::MatrixBase<derived_type>::ColsAtCompileTime ==
+               1) ALGEBRA_HOST_DEVICE
+      static inline auto rotate(
+          const Eigen::Transform<scalar_type, 3, Eigen::Affine> &m,
+          const Eigen::MatrixBase<derived_type> &v) {
 
     return m.matrix().template block<3, 3>(0, 0) * v;
   }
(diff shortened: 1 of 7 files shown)


math/eigen/include/algebra/math/impl/eigen_transform3.hpp
```diff
@@ -81,6 +79,8 @@ struct transform3 {
   ALGEBRA_HOST_DEVICE
   transform3(const vector3 &t, const vector3 &x, const vector3 &y,
              const vector3 &z, bool get_inverse = true) {
+    _data.setIdentity();
```
beomki-yeo, line 82: I think you also need to set `_data_inv` to identity matrix

storage/common/include/algebra/storage/matrix_getter.hpp
```diff
@@ -241,7 +241,7 @@ ALGEBRA_HOST_DEVICE constexpr void set_block(
   if constexpr (ROWS == mROW &&
                 matrix_t::storage_rows() == input_matrix_t::storage_rows()) {
     if (row == 0u) {
-      for (std::size_t j = col; j < mCOL; ++j) {
+      for (std::size_t j = col; j < col + COLS; ++j) {
```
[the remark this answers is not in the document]
niermann999: Not sure. The vector getter used ```SubCol``` previously, which worked in detray as long as we used it with ```row=0```. With any other row index, it starts to go out of bounds. So, I put in this code and made a mistake myself

math/eigen/include/algebra/math/impl/eigen_transform3.hpp
niermann999: Ah, right, I keep forgetting that ```get_inverse``` could be ```false```. At that point, accessing ```_data_inv``` should not be permitted actually, regardless if it is the identity matrix or garbage

storage/common/include/algebra/storage/matrix_getter.hpp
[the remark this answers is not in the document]
niermann999: Ah, wait, this is the matrix getter for SoA types. This is new code

math/eigen/include/algebra/math/impl/eigen_transform3.hpp
niermann999: Done

beomki-yeo (approved):

krasznaa (approved):
Apart from the formatting issues, this looks fine to me.

Just as a comment: Instead of the `using namespace algebra` statements, I myself would've probably put the common test code into let's say an `algebra::test::` namespace.

But since it's testing code, I don't actually care a whole lot about the exact formalism...

PR #144, opened 2024-12-18
fix: benchmark deprecation warning

niermann999:
Try to fix google benchmark deprecation warning. Also uses explicit return types from the benchmark cases, to prevent spurious optimizations


beomki-yeo (approved):
lgtm

PR #145, opened 2025-02-19
Fix debug compilation with HIP

StewMH:
Add HIP runtime header for assert when compiling as a HIP source file: avoid host/device function issues.

### DIFF
--- a/storage/cmath/include/algebra/storage/impl/cmath_getter.hpp
+++ b/storage/cmath/include/algebra/storage/impl/cmath_getter.hpp
@@ -17,6 +17,10 @@
 #include <cstddef>
 #include <type_traits>
 
+#if defined(__HIP__)
+#include <hip/hip_runtime.h>
+#endif
+
 namespace algebra::cmath::storage {
 
 /// "Element getter", assuming a simple 2D array access


StewMH:
Actually this only solves one of the instances of assert, which is enough for the current traccc alpaka hip implementation. If we don't want to replicate this ifdef across a lot of files we probably need a dedicated header e.g. `#include <device_assert.hpp>`?

niermann999:
> Actually this only solves one of the instances of assert, which is enough for the current traccc alpaka hip implementation. If we don't want to replicate this ifdef across a lot of files we probably need a dedicated header e.g. `#include <device_assert.hpp>`?

Yes, I think such a header would be good, don't know what other peoples opinions are

StewMH:
I haven't managed to provoke more of these errors in my tests: maybe we just go ahead with this and revisit if more instances are needed.

niermann999:
> I haven't managed to provoke more of these errors in my tests: maybe we just go ahead with this and revisit if more instances are needed.

Sorry for the late reply! Even if we use it only in one place, could we still put it in a dedicated header? I think that would make it both more readable and extendible in future. E.g. `common/assert.hpp` or something similar
```
/**
 * ALGEBRA PLUGIN library, part of the ACTS project (R&D line)
 *
 * (c) 2025 CERN for the benefit of the ACTS project
 *
 * Mozilla Public License Version 2.0
 */

#pragma once

#include <cassert>

#if defined(__HIP__)
#include <hip/hip_runtime.h>
#endif
```

PR #146, opened 2025-02-28
Simplify dependency structure

stephenswat:
This commit significantly simplifies the dependency structure by removing the backends for vecmem and smatrix, as well as frontends which combine non-matching frontends and backends. This is functionality we never ever use and it is simply complicating this repository.


tests/accelerator/cuda/common/test_cuda_basics.cuh
```diff
@@ -26,11 +24,63 @@ class test_cuda_basics : public test_basics_base<T> {
...
+  test_cuda_basics() : test_basics_base<T>() {}
 
  protected:
-  /// Memory resource for all of the tests.
-  vecmem::cuda::managed_memory_resource m_resource;
+  /// Function setting things up for a test
+  virtual void SetUp() override {
+
+    // Create the vectors.
+    CUDA_ERROR_CHECK(cudaMallocManaged(
```
krasznaa, line 34: I am **not** on board with this.

I agree with getting rid of the "vecmem frontend" of the code. But we still shall continue to use vecmem to test the code. Since we're using vecmem for the memory management of everything in all the downstream projects as well. So we should test the synergy between these codebases already here.

I.e. the project will definitely need to continue building vecmem for its tests.

Note that once we merge the project into detray, this is not "anything" even. Since detray is not going to stop depending on vecmem either.

frontend/eigen_generic/include/algebra/eigen_generic.hpp
```diff
@@ -1,127 +0,0 @@
...
- *
- * (c) 2020-2024 CERN for the benefit of the ACTS project
- *
- * Mozilla Public License Version 2.0
- */
-
-#pragma once
-
-// Project include(s).
-#include "algebra/math/eigen.hpp"
```
krasznaa, line 11: You were a bit over-eager with this one as well I fear. 🤔

Let's discuss which frontends we want to get rid of. I don't want to throw out the baby as well in all of this...

tests/accelerator/sycl/CMakeLists.txt
```diff
@@ -4,34 +4,29 @@
...
 add_library( algebra_tests_sycl_common INTERFACE )
 target_include_directories( algebra_tests_sycl_common INTERFACE
    "${CMAKE_CURRENT_SOURCE_DIR}/common" )
 target_link_libraries( algebra_tests_sycl_common
-   INTERFACE GTest::gtest algebra::tests_common algebra::tests_accel_common
-             vecmem::core vecmem::sycl )
+   INTERFACE GTest::gtest algebra::tests_common algebra::tests_accel_common )
 add_library( algebra::tests_sycl_common ALIAS algebra_tests_sycl_common )
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsycl")
```
krasznaa, line 15: Come on... No. This is not flying like this. Why did you think that hardcoding the use of just oneAPI like this is the right move? 😕

stephenswat:
> This has been (partially) superseded by #168 , which prepares for the detray integration. Do you mind closing it or is there anything in these changes that should still go in?

Thanks!

PR #148, opened 2025-03-11
feat: Cleanup benchmark names and add inversion and determinant

niermann999:
Uses constexpr string views for the benchmarks names now and adds the new cases for the determinant and matrix inversion that were missing previously. The Eigen plugin inversion algorithm is returning a concrete type now in order to prevent expression templates to be handed through.


beomki-yeo (approved):
Looks good to me

PR #149, opened 2025-03-11
feat: loop unrolling and switch to constexpr

niermann999:
Forces the compiler to apply loop unrolling in the Vc plugins, which seems to help performance (the array plugin seems to be optimized that way already). For this, a new macro was added to the qualifiers header to do this per compiler (however, I could not find anything for MSVC so far). I think it is OK to mandate loop unrolling at some loops that seem to make sense, since we don't have large matrices to run

Since `constexpr` implies `inline` most functions were switched to `constexpr` now.

### DIFF
--- a/common/include/algebra/qualifiers.hpp
+++ b/common/include/algebra/qualifiers.hpp
@@ -31,3 +31,14 @@
 #else
 #define ALGEBRA_ALIGN(x) alignas(x)
 #endif
+
+// @see https://stackoverflow.com/questions/78071873/gcc-preprocessor-macro-and-pragma-gcc-unroll
+#if defined(__clang__)
+#define ARG_TO_STRING(A) #A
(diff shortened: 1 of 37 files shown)


common/include/algebra/qualifiers.hpp
```diff
@@ -31,3 +31,21 @@
...
+#if __GNUC__ >= 14
+#define ALGEBRA_UNROLL_N(n) _Pragma(ARG_TO_STRING(GCC unroll n))
+#else
+// For versions below 14, template parameters apparently cannot be used
+#define ALGEBRA_UNROLL_N(n) _Pragma(ARG_TO_STRING(GCC unroll 8))
+#endif
+#else
+// Unknown compiler or does not support unrolling directives
+#define ALGEBRA_UNROLL_N(n)
+#endif
```
stephenswat, line 51: Clever!

stephenswat (approved):

PR #154, opened 2025-03-13
Minor fixes and improvements

wermos:
I updated the README to reflect the `ALGEBRA_PLUGINS_BUILD_BENCHMARKS` CMake option.

A bunch of the benchmarking code source files either used `#include <string>` or had no `#include` whatsoever, but was using `std::cout`. I changed all of those to `#include <iostream>`.

Half of the Eigen benchmarks called the library "Eigen3" and the other half called it "Eigen". I changed it so that all of them just say "Eigen".


niermann999 (approved):
Thanks!

PR #155, opened 2025-03-17
feat: vectorized implementation of the cross poduct

niermann999:
Use shuffle operations for the implementation of the cross product in the Vc AoS plugin. Since there is a trailing zero element to reach a fixed alignment, this has to be initialized to a specific value from the input vectors, so that the shuffles result in the correct vectors to do the multiplication and subtraction with.


storage/vc_aos/include/algebra/storage/vc_aos.hpp
```diff
@@ -90,6 +90,16 @@ struct index<Vc_1::Vector<T, Vc_1::simd_abi::fixed_size<N>>> {
...
 
+template <typename T, auto N>
+struct value<vc_aos::storage_type<T, N>> {
+  using type = T;
+};
+
+template <typename T, auto N>
+struct value<Vc_1::Vector<T, Vc_1::simd_abi::fixed_size<N>>> {
+  using type = T;
+};
```
stephenswat, line 101: Would it make sense to template this on a typename instead of an integer `N`?

stephenswat, line 101: Would it make sense to template this on a typename instead of an integer `N`?

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
```diff
@@ -102,8 +136,11 @@ ALGEBRA_HOST_DEVICE constexpr auto eta(const vector_t &v) noexcept {
 /// @return a vector representing the cross product
 template <algebra::concepts::vc_aos_vector vector_t1,
           algebra::concepts::vc_aos_vector vector_t2>
-ALGEBRA_HOST_DEVICE constexpr auto cross(const vector_t1 &a, const vector_t2 &b)
-    -> decltype(a * b - b * a) {
+requires(std::same_as<algebra::traits::value_t<vector_t1>, double> ||
+         std::same_as<algebra::traits::value_t<vector_t2>, double>)
```
stephenswat, line 140: Shouldn't they both be doubles?

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
```diff
@@ -88,11 +85,48 @@ ALGEBRA_HOST_DEVICE constexpr auto normalize(const vector_t &v) {
...
+/// @tparam vector_t generic input vector type
+///
+/// @param a the first input vector
+/// @param b the second input vector
+///
+/// @return a vector representing the cross product
+template <algebra::concepts::vc_aos_vector vector_t1,
+          algebra::concepts::vc_aos_vector vector_t2>
+requires(std::same_as<algebra::traits::value_t<vector_t1>, float>
+             &&std::same_as<algebra::traits::value_t<vector_t2>, float>)
```
stephenswat, line 102: Why do we need two separate use cases for float and double?

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
niermann999: Indeed, I don't think we ever have a mixed precision case, but I thought it was safer to use the "negation" of the single precision case instead of potentially having no overload that fits

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
niermann999: The float implementation compiles into what looks like inefficient code for double precision - as far as I can tell. It seems to move things from avx registers to sse registers, then doing the rotation, then back etc. The implementation with fma looks much better on double precision vectors (and seems to be about 2-3 times faster in the benchmark)

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
stephenswat: Ah, right. But can float and double vectors be crossed with eachother?

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
stephenswat: Very strange, but I'm very happy to accept this as fact.

storage/vc_aos/include/algebra/storage/vc_aos.hpp
niermann999: You mean the `value` struct? The other type `Vc_1::Vector<T, Vc_1::simd_abi::fixed_size<N>>` is not not part of algebra plugins

stephenswat: No what I mean is to replace this by:

```suggestion
template <typename T, typename Q>
struct value<Vc_1::Vector<T, Q>> {
  using type = T;
};
```

math/vc_aos/include/algebra/math/impl/vc_aos_vector.hpp
niermann999: I guess we would see a bunch of double to single conversion warnings (as it probably should be atm). But the double implementation just rips every element of the simd vector out and calculates the cross product "the old fashioned way". So, I guess it should technically work

stephenswat: Fair enough!

storage/vc_aos/include/algebra/storage/vc_aos.hpp
niermann999: Ah, Ok! Makes sense, let me try

stephenswat (approved):

storage/vc_aos/include/algebra/storage/vc_aos.hpp
niermann999: My guess is, that fixed size is the abi for the SimdArray (which has a fixed number of elements rather than a fixed vector register type). Using it might be more explicit compared to the SoA plugin

PR #157, opened 2025-03-17
ref: Update externals

niermann999:
Update Fastor to version 0.6.4
Update google test to version 1.16.0
Update Vc to version 1.4.5
Update vecmem to version 1.14.0


niermann999:
I get some very different results in the benchmarks now, I need to have another look

niermann999:
I think the changes in the reported benchmark timings come from this issue https://github.com/google/benchmark/issues/1834 with this fix https://github.com/google/benchmark/pull/1836, since dividing the new reported time by the number of threads gives roughly the old reported time

PR #159, opened 2025-03-19
feat: Algebra type casts

niermann999:
Implement functionality to cast algebra types to different floating point precision. This will be helpful in testing where floating point errors come from in downstream projects.

Edit: Also applied a sonar cloud suggestion on member initialization in the fastor transform type

### DIFF
--- a/math/eigen/include/algebra/math/impl/eigen_transform3.hpp
+++ b/math/eigen/include/algebra/math/impl/eigen_transform3.hpp
@@ -60,6 +60,10 @@ struct transform3 {
   using matrix44 =
       typename Eigen::Transform<scalar_type, 3, Eigen::Affine>::MatrixType;
 
+  /// Helper type to cast this to another floating point precision
+  template <concepts::scalar o_scalar_t>
+  using other_type = transform3<o_scalar_t>;
+
   /// @}
 
   /// @name Data objects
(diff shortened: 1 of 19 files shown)


math/eigen/include/algebra/math/impl/eigen_transform3.hpp
```diff
@@ -128,6 +132,14 @@ struct transform3 {
     _data_inv = _data.inverse();
   }
 
+  /// Constructor with arguments: matrix and its inverse
+  ///
+  /// @param m is the full 4x4 matrix
+  /// @param m_inv is the inverse to m
+  ALGEBRA_HOST_DEVICE
+  transform3(const matrix44 &m, const matrix44 &m_inv)
+      : _data{m}, _data_inv{m_inv} {}
```
stephenswat, line 141: Would it make sense to add a debug-mode assertion that `m_inv` is indeed the inverse if `m`?

math/fastor/include/algebra/math/impl/fastor_transform3.hpp
```diff
@@ -115,20 +119,24 @@ struct transform3 {
...
     _data_inv = Fastor::inverse(_data);
   }
 
+  /// Constructor with arguments: matrix and its inverse
+  ///
+  /// @param m is the full 4x4 matrix
+  /// @param m_inv is the inverse to m
+  ALGEBRA_HOST
+  transform3(const matrix44 &m, const matrix44 &m_inv)
+      : _data{m}, _data_inv{m_inv} {}
```
stephenswat, line 132: See above.

stephenswat (approved):

math/eigen/include/algebra/math/impl/eigen_transform3.hpp
niermann999: Done

math/fastor/include/algebra/math/impl/fastor_transform3.hpp
niermann999: Done

PR #162, opened 2025-04-29
Add function definitions for function calls involving intermediate Fastor types

wermos:


wermos:
cc @niermann999 for review. This mostly is in response to what we discussed in [detray PR #963](https://github.com/acts-project/detray/pull/963).

niermann999:
> cc @niermann999 for review. This mostly is in response to what we discussed in [detray PR #963](https://github.com/acts-project/detray/pull/963).
> 
> I've added definitions for intermediate Fastor types for `dot` and `cross` so far; I also need to add one for `normalize` for the `detray` Fastor build to run.
> 
> Ideally though I should add these function definitions for all the functions, I think. What are your thoughts on the matter?

Yes, if this works, I think it should be done for all functions. Also, is it possible to remove the overloads using Fastor tensors?

wermos:
I've rewritten the `phi`, `theta`, `perp`, `norm`, `eta`, and `normalize` functions so that they play nicely with intermediate Fastor types.

The set of overloads for `dot` and `cross` is the minimal set required to get the tests and benchmarks to compile. Removing any of the overloads causes compilation errors.

For both `dot` and `cross`, 4 overloads are required to cover all the cases:
1. Expression template (`Fastor::AbstractTensor<Derived, N>`) with expression template (`Fastor::AbstractTensor<Derived, N>`).
2. Pure vector (`Fastor::Tensor<T, N>`) with pure vector (`Fastor::Tensor<T, N>`).
3. Pure vector (`Fastor::Tensor<T, N>`) with matrix (slice) (`Fastor::Tensor<T, N, 1>`).
4. Matrix (slice) (`Fastor::Tensor<T, N, 1>`) with pure vector (`Fastor::Tensor<T, N>`).

wermos:
Also, in order to support the intermediate types, I had to remove the `requires` clauses. It's not possible to determine if the type an `AbstractTensor` contains inside it has the required shape, at compile-time.

wermos:
@niermann999 This PR is ready for review now.

PR #163, opened 2025-06-19
Implement transposed matrix multiplication

stephenswat:
This commit adds a new `transposed_product` method which allows for matrix multiplication of matrices with either the left, the right, or both matrices transposed without the need to construct any intermediate matrices.

### DIFF
--- a/math/generic/include/algebra/math/impl/generic_matrix.hpp
+++ b/math/generic/include/algebra/math/impl/generic_matrix.hpp
@@ -273,6 +273,56 @@ set_inplace_product_left_transpose(MA &A, const MB &B) requires(
   }
 }
 
+template <bool transpose, concepts::matrix M>
+ALGEBRA_HOST_DEVICE constexpr algebra::traits::value_t<M> transposable_get(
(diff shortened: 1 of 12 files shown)


math/generic/include/algebra/math/impl/generic_matrix.hpp
```diff
@@ -273,6 +273,62 @@ set_inplace_product_left_transpose(MA &A, const MB &B) requires(
...
+  using effective_rhs_t =
+      std::conditional_t<transpose_right,
+                         decltype(transpose(std::declval<MB>())), MB>;
+
+  using result_t =
+      algebra::traits::get_matrix_t<MA, algebra::traits::rows<effective_lhs_t>,
+                                    algebra::traits::columns<effective_rhs_t>,
+                                    value_t>;
+
+  result_t C;
```
niermann999, line 311:
```suggestion
  auto C = zero<result_t>();
```
Or is this slow, because it is another assignment, temporary etc?

niermann999 (approved):
How does this fit in with the `generic::math::set_inplace_product_left_transpose` and the other methods like it?

